PR #2, opened 2021-02-05
Add Author view

jmmxp:
This view allows us to GET all existing Authors and also POST new ones. Will need to look into how to provide an endpoint to only get info for one Author.
image


VinnyLuu (approved):
LGTM

PR #3, opened 2021-02-06
Add GET author by _id tests

jmmxp:
I've added one test to get a valid author (should return 200 OK), and ensure the JSON is correct, and one test for an invalid author (should return 404 Not Found).


VinnyLuu (approved):
LGTM

PR #4, opened 2021-02-06
Implemented Post and getting of post with endpoint /posts/_id

VinnyLuu:
Set up Post model and getting of post with endpoint /posts/_id


back-end/api/quickstart/models.py
```diff
@@ -1,13 +1,43 @@
...
+  _id = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  _type = 'post'
+  title = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  description = models.TextField()
+  source = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  origin = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  visibility = models.CharField(max_length=SHORT_CHAR_LENGTH, choices=Visibility.choices)
+  unlisted = models.BooleanField()
+  isPrivateToFriends = models.BooleanField()
+  author = models.JSONField()
```
jmmxp: Should author just be an author ID (URL) instead? It seems like you're treating it as an actual author object, but we might have to get authors from other sites right?

back-end/api/quickstart/models.py
```diff
@@ -1,13 +1,43 @@
...
+  contentType = models.CharField(max_length=LONG_CHAR_LENGTH)
+  content = models.TextField(blank=True)
+  categories = models.JSONField()
+  # cannot store time as IOS 8601 as per spec so when interacting will need parse (i.e using django.utils.dateparse)
+  published = models.TimeField(default=timezone.now)
+  count = models.IntegerField()
+  pageSize = models.IntegerField(default=50)
+  commentLink = models.CharField(max_length=LONG_CHAR_LENGTH)
+  comments = models.JSONField()
```
jmmxp, line 44: Same as above

back-end/api/quickstart/tests.py
```diff
@@ -27,3 +29,46 @@ def test_get_valid_author(self):
...
+    self.post = Post.objects.create(
+      _id='123',
+      title='testpost',
+      description='i am a test post',
+      source='source post id',
+      origin='origin post id',
+      visibility='Public',
+      unlisted='True',
+      isPrivateToFriends='True',
+      author=AuthorSerializer(self.john).data,
```
jmmxp: If you implement the changes I talked about above, the test might get a bit difficult (you'd have to do author=testId here, but then it's not so easy to make assertions on the HTTP response anymore). Maybe we leave it for a future PR?

jmmxp (changes requested):
LMK what you think re; author/comment fields

back-end/api/quickstart/views.py
```diff
@@ -25,8 +25,19 @@ class GroupViewSet(viewsets.ModelViewSet):
 
 class AuthorViewSet(viewsets.ModelViewSet):
     """
-    API endpoint that allows groups to be viewed or edited.
+    API endpoint that allows authors to be viewed or edited.
     """
     queryset = Author.objects.all()
     serializer_class = AuthorSerializer
+    # Using lookup_field as search param
+    # https: // stackoverflow.com / questions / 56431755 / django - rest - framework - urls - without - pk
```
jmmxp: Redundant spacing

back-end/api/quickstart/models.py
VinnyLuu: Fixed

back-end/api/quickstart/models.py
jmmxp: We discussed this offline - this should stay as JSONField because it should be an array of URLs.

jmmxp (approved):
LGTM

back-end/api/quickstart/tests.py
VinnyLuu: The assertions will pass now, just later on when we have to actually return the json object will these test not pass but that can be dealt then.

PR #17, opened 2021-02-06
Add update author test

jmmxp:
Test uses PUT as is, may need to be changed to POST in the future. Should be as easy as changing `.put` to `.post`.

### DIFF
--- a/back-end/api/quickstart/tests.py
+++ b/back-end/api/quickstart/tests.py
@@ -66,6 +66,18 @@ class UpdateAuthorById(TestCase):
     for k in self.payload:
       self.assertEqual(serializer.data[k], self.payload[k])
 
+    # Ensure other fields are unchanged
+    self.assertEqual(serializer.data['_id'], self.john._id)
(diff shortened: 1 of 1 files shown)


back-end/api/quickstart/tests.py
```diff
@@ -31,6 +32,36 @@ def test_get_invalid_author(self):
...
+
+  # TODO: update object tests most likely will be changed to use POST instead.
+  def test_update_author(self):
+    response = client.put(
+      f'/authors/{self.john._id}/',
+      data=json.dumps(self.payload),
+      content_type='application/json'
+    )
+    self.assertEqual(response.status_code, status.HTTP_200_OK)
+
```
VinnyLuu, line 63: Should you get the author here and assert if its actually updated?

jmmxp: Done.

jmmxp: Now checking that the PUT'ed new fields are updated, and the rest are unchanged.

VinnyLuu (approved):
LGTM

PR #18, opened 2021-02-06
Add update valid/invalid post tests

jmmxp:
Two new tests: update_post (expects a 200 OK, and then checks the Post's data has actually changed), and update_invalid post which expects a 404.


back-end/api/quickstart/tests.py
```diff
@@ -112,3 +112,76 @@ def test_get_valid_post(self):
   def test_get_invalid_post(self):
     response = client.get('/posts/invalidId/')
     self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+
+class UpdatePostById(TestCase):
+  """Tests for updating a single Author by PUT'ing to endpoint /author/_id/.
```
VinnyLuu: posts instead of author

back-end/api/quickstart/tests.py
```diff
@@ -112,3 +112,76 @@ def test_get_valid_post(self):
...
+      # TODO: Remove _id here once we figure out the _id field. It should be uneditable.
+      '_id': '123', 
+      'title': 'newpost', 
+      'description': 'i am a new post', 
+      'source': 'new source post id', 
+      'origin': 'new origin post id', 
+      'visibility': 'F', 
+      'unlisted': False, 
+      'isPrivateToFriends': False, 
+      'author': 'newId',
```
VinnyLuu, line 151: this author id should also be uneditable like the _id right? cause it shouldn't change when updating posts

VinnyLuu (approved):
LGTM

back-end/api/quickstart/tests.py
jmmxp: Good point, added a comment here to remove this later as well (once fields are figured out).

PR #20, opened 2021-02-07
Allow POST for update instead of PUT, fix tests, rename endpoints

jmmxp:
Instead of using routers, binding views to endpoints specifically now, which allows us to explicitly list that POST should be used for updates instead of PUT.

Changed the `/authors/{AUTHOR_ID}/` endpoint to `/author/{AUTHOR_ID}/`, and changed `/posts/{POST_ID}/` to `/author/{AUTHOR_ID}/posts/{POST_ID}/`, as per the spec.

Tests updated to POST instead of PUT, and also endpoints were changed accordingly (as well as comments).


VinnyLuu (approved):
LGTM

PR #22, opened 2021-02-07
Add test for PUT'ing a new Post

jmmxp:
The test only checks for a 201 CREATED response code - there's no need to test any further as of now, because the current PUT functionality is given to us for free from using ModelViewSet with Django Rest Framework.

The test will have to be modified/extended upon if we ever have to override the creation functionality, though.


VinnyLuu (approved):
LGTM

PR #23, opened 2021-02-07
Add tests for DELETE'ing a Post

jmmxp:
Two tests: one that sends a valid DELETE request and ensures the response code is 204 No Content, and another that sends an invalid DELETE request and ensures the response code is 404 Not Found.


VinnyLuu (approved):
LGTM

PR #24, opened 2021-02-07
Clean tests up by creating helper to return test object data

jmmxp:
As suggested by Vinny, I've added helpers to return test data for both test_posts and test_author.


back-end/api/quickstart/test_posts.py
```diff
@@ -9,27 +9,32 @@
...
+    'origin': 'origin post id',
+    'visibility': 'Public',
+    'unlisted': 'True',
+    'isPrivateToFriends': 'True',
+    'author': 'testId',
+    'contentType': 'text/plain',
+    'content': 'Hello, I am a test post',
+    'categories': '["Testing"]',
+    'published': datetime.datetime.now(),
+    'count': '5',
```
VinnyLuu: 5

VinnyLuu (approved):

back-end/api/quickstart/test_posts.py
jmmxp: Fixed this and the booleans as well

PR #28, opened 2021-02-08
Add followers list endpoint

jmmxp:
`/author/{AUTHOR_ID}/followers` endpoint has now been added with a working test.

I had to override the list() method within the ViewSet, and get the list of Author objects myself. Note that we'll most likely have to change it to make API calls instead of getting Authors from database.


VinnyLuu (approved):
LGTM

PR #34, opened 2021-02-16
Add Like model, GET endpoints and tests for likes of a post or comment

jmmxp:
Like model has been added.

GETs to `api/author/<str:author>/posts/<str:post>/likes/` and  `api/author/<str:author>/posts/<str:post>/comments/<str:comment>/likes` are now implemented. 

The behaviour for both is exactly the same as of now - it filters Like objects with the given post ID OR comment ID but this may change if we need to use the author or author/post parameters in the URL.


back-end/api/api/urls.py
```diff
@@ -9,6 +9,7 @@
 router.register(r'authors', views.AuthorViewSet)
 router.register(r'posts', views.PostViewSet)
 router.register(r'followers', views.FollowersViewSet)
+router.register(r'likes', views.LikesPostViewSet)
```
VinnyLuu, line 15: missed LikesCommentViewSet?

back-end/api/quickstart/test_likes_comment.py
```diff
@@ -0,0 +1,38 @@
...
+  """Tests for getting all likes of a comment at endpoint api/author/<str:author>/posts/<str:post>/comments/<str:comment>/likes."""
+  def setUp(self):
+    self.test_comment_id = 1
+    self.like1 = Like.objects.create(**get_test_like_fields(comment_id=self.test_comment_id))
+    self.like2 = Like.objects.create(**get_test_like_fields(comment_id=self.test_comment_id))
+    self.like3 = Like.objects.create(**get_test_like_fields(comment_id=2))
+
+    self.likes = [self.like1, self.like2]
+
+  def test_get_likes_for_post(self):
```
VinnyLuu: test_get_likes_for_comment

VinnyLuu (changes requested):

back-end/api/api/urls.py
jmmxp: No, only need to register it for one. They both present the same thing (likes). I added a comment about this.

VinnyLuu (approved):
LGTM

PR #35, opened 2021-02-16
Fix debug site for comments

jmmxp:
image

We do this by using retrieve instead of list in ViewSet. Whenever list is overridden, the debug site page we get from registering with the router does not work.

We can decide later if we want to keep it this way or change it back to retrieve once we've made more progress.


VinnyLuu (approved):
LGTM. Should this also be done to the post list as well?

jmmxp:
No need to do it to the post list, the reason why that one works is because we have a Singular/Detail view and a separate List View, and we registered the router with the Singular view, which doesn't override 'list'.

PR #36, opened 2021-02-16
Add UUIDs to URLs and IDs

### LINKED ISSUE #30: Save Objects to Database with UUIDs
When we save objects to the database, `save()` is called. We can customize this method so that whenever we post or put anything with an ID, we can save UUIDs to be part of the URL. Need to look into how to do this properly.

For example, when creating a new author, we can customize/override the `save()` method to append a UUID string to the URL when we put/post. Django should know the difference between updating and creating a new author so a new UUID string doesn't always get generated on save.

https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.save

raysarinas:
Used `uuid.uuid4().hex` to get rid of the hyphens

Closes #30


back-end/api/quickstart/signals.py
```diff
@@ -7,6 +7,11 @@
...
 def after_user_signed_up(request, user, **kwargs):
-    #TODO: Make this url a uuid
-    url = user.id
+    """
+    After a user has signed up, an author is automatically
+    created with the appropriate id containing a UUID
+    Get absolute URI: https://stackoverflow.com/a/10119243
+    """
+    hex_uuid = str(uuid.uuid4().hex)
+    url = str(request.build_absolute_uri('/')) + "author/" + hex_uuid
```
jmmxp, line 16: Just want to confirm that this url contains the port of the server (and not the client) like Chris was concerned about.

back-end/api/quickstart/migrations/0020_auto_20210215_2314.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('quickstart', '0019_auto_20210215_2214'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='author',
+            name='_id',
+            field=models.CharField(max_length=128, primary_key=True, serialize=False),
```
jmmxp, line 16: Is the PR missing a migration? It seems that the newest changes to the models don't include your change to the author model as defined in this migration.

jmmxp (approved):
LGTM just some comments to address

Also remember to run ./manage.py tests before submitting

back-end/api/quickstart/signals.py
raysarinas: yes, from what I tested the urls returned are `http://localhost:8000/author/author_id`

back-end/api/quickstart/migrations/0020_auto_20210215_2314.py
raysarinas: what change to the author model? I don't think I changed anything. The changes I made are in `signals.py` but didn't cause any migrations

PR #38, opened 2021-02-16
Refactor test data methods to helper

jmmxp:
Prereq work before I add inbox tests, otherwise the tests will end up with a lot of duplicate code.


VinnyLuu (approved):
LGTM

PR #39, opened 2021-02-16
Implement inbox endpoint

VinnyLuu:
@Jmmxp and I worked together to implement the inbox endpoint. Can send GET, POST and DELETE to /api/author/<str:author>/inbox/

Tests coming in an upcoming PR.


back-end/api/api/urls.py
```diff
@@ -61,6 +62,11 @@
     'get': 'retrieve'
 })
 
+inbox = views.InboxViewSet.as_view({
+    'get': 'retrieve',
+    'post': 'update',
+    'delete': 'clear'
```
jmmxp: Should this be 'delete': 'destroy' instead? I think that's the method DRF viewsets use.

jmmxp (approved):
LGTM

PR #41, opened 2021-02-17
Created new inbox for author on registration

VinnyLuu:
Hooked into the signals.py file and added new inbox for the new author on registration

Testing: Ran locally and registered through the front end, checking in the backend the inbox was created


jmmxp (approved):
LGTM

PR #49, opened 2021-02-19
Create posts list, as well as a posts detail modal

ChrisChrisLoLo:
Create list view of posts on the home page.

Create a detail view for the post.

Added a new Post type.

The added TSX is currently bare-bones so that it should be easily adaptable when the API is finalized.


front-end/src/pages/HomePage.tsx
```diff
@@ -19,6 +32,19 @@ export default function HomePage(props:RouteComponentProps) {
...
+  if (postEntries === undefined){
+    postListElToDisplay = <p>Loading...</p>;
+  } else if (postEntries.length === 0) {
+    postListElToDisplay = <p>No Entries Found</p>;
+  } else {
+    postListElToDisplay = postEntries.map((post:Post)=>
+      <PostListItem post={post}/>
+    );
+  }
+  console.log(postListElToDisplay)
```
VinnyLuu, line 46: remove console.log

front-end/src/types/Post.ts
```diff
@@ -0,0 +1,19 @@
+export interface Post {
+  _id: string
+  _type: string
+  title: string
+  description: string
+  source: string
+  origin: string
+  visibility: string
+  unlisted: boolean
+  isPrivateToFriends: boolean
```
VinnyLuu: TODO: probably gonna remove this when we fix up on the models in the backend since I accidentally left it in

front-end/src/components/PostListItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const toggle = () => setIsModalOpen(!isModalOpen);
+
+  const post: Post = props.post;
+  return (
+    <div>
+      <Card>
+        {/* TODO: Post image here? */}
```
VinnyLuu, line 20: Yea this is something we might need to discuss about. I think Hindle talked about image post being seperate? so it might complicate things but we'll see :)

VinnyLuu (approved):
LGTM

front-end/src/pages/HomePage.tsx
jmmxp: Let's actually leave these here for now, I though the same thing at first but IMO they should be really useful for debugging

We can ask the TA if they care about debugging statements left in the code this week and remove them before the deliverables if they do care.

front-end/src/types/Post.ts
jmmxp: @VinnyLuu Could you remove that from the backend in a PR now so Chris can just remove isPrivateToFriends?

VinnyLuu: Sure I'll get it up rn

front-end/src/pages/HomePage.tsx
VinnyLuu: Sounds good

front-end/src/components/PostListItem.tsx
ChrisChrisLoLo: Yeah definitely! What I was thinking is adding logic to show either the image HTML or the text HTML, depending on the content type. I might be misunderstanding the spec though, so talking about this would be good!

front-end/src/pages/HomePage.tsx
ChrisChrisLoLo: Yeah my bad, bad habit of mine 😅. I'm fine with either way. It might be more useful to leave them in for people new to the codebase, but I should try to limit the amount of white noise going to the logs going forward

PR #52, opened 2021-02-20
Display image for image posts (detail view)

jmmxp:
I check contentType and if it's any of the image types, I use an `img` tag instead of a `p` tag to display the content.

Also made sure to add 'content' as a string field on the Post class, something we missed earlier.

image

### DIFF
--- a/front-end/src/components/PostDetailModal.tsx
+++ b/front-end/src/components/PostDetailModal.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
 import { Post } from '../types/Post';
+import { isImage } from '../helpers/ImageHelper';
 
 interface Props {
   post: Post;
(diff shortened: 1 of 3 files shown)


front-end/src/components/PostDetailModal.tsx
```diff
@@ -10,13 +11,23 @@ interface Props {
 
 export default function PostListItem(props:Props) {
   const post: Post = props.post;
+
+  function Content() {
```
ChrisChrisLoLo, line 15: I think it's pretty clever to use functions like this over what I'm used to, which is just changing around scoped variables 👍

ChrisChrisLoLo (approved):
LGTM!

Does the backend currently serialize `content`, or does that need to be added in another PR?

front-end/src/components/PostDetailModal.tsx
jmmxp: Good to know! I feel like I have no idea what I'm doing when it comes to React haha

jmmxp:
It's already being serialized, so we're good to go :)

ChrisChrisLoLo:
Cool. Good catch, I must have missed that field in my last PR

PR #53, opened 2021-02-20
Fix broken test, change author fields to be named id and type

jmmxp:
Test was broken from old test that still used isPrivateToFriends, I've removed that helper function since it's not actually needed.

Changed our model to use `id` and `type` internally.


VinnyLuu (approved):
LGTM

PR #54, opened 2021-02-21
Rename type and object on Like model, Inbox

jmmxp:
Cleaning up the models to match the spec.

Renamed _type and _object to type and object for the Like Model. 
Renamed _type to type for Inbox and corrected the value to 'inbox'.


raysarinas:
LGTM

raysarinas (approved):

PR #55, opened 2021-02-21
Rename Comment model fields

jmmxp:
`_id` -> `id` and `_type` -> `type`. Fixed up tests as well to accommodate this change.

I also removed postId from being serialized - this should be a Foreign Key to a post and never returned to the client.


back-end/api/quickstart/helper_test.py
```diff
@@ -33,8 +33,7 @@ def get_test_like_comment_fields(comment_id=1):
 
 def get_test_comment_fields(i=1, postid="testpostid"):
```
VinnyLuu: postid not used anymore so you can take it out

VinnyLuu (changes requested):

back-end/api/quickstart/helper_test.py
jmmxp: Done

VinnyLuu (approved):
LGTM

PR #56, opened 2021-02-21
Rename Post model fields

jmmxp:
Updated them to match the spec, `_type` has been renamed to `type` and `_id` has been renamed to `id`.

`id` has also been changed to a primary key.


Monica-Uwalaka (approved):
This looks better

PR #57, opened 2021-02-21
Author settings page

### LINKED ISSUE #47: UI for Author Settings
### Related User Stories
**24. As an author, I want to be able to use my web-browser to manage my profile. **

### Description / Notes
- UI for displaying author information that can be edited/updated (POST an author)
- This task could be limited to updating info about an author (git URL, display name)
- May involve dealing with user authentication - but this can be resolved in another issue

### Tasks
- [ ] UI for editing author profile information
- [ ] Test being able to send a POST request and update as appropriate

raysarinas:
Closes #47 

Settings page where user's can change their display name and GitHub url.

As of right now, we are querying for all users, then finding a user's username == author's display name, then updating both display name and/or GitHub url.

I suspect we will run into pagination issues as GET results are returned in pages. As the number of users grows, we then need to iterate through all of them to find a matching author display name. Would prefer if there could be a way around this since the only solution to this problem I can think of is having a while loop to iterate through each page of users/authors. 

Another issue we will run into by having users and authors "connected" to one another via display name is that when an author's display name is changed, there's no way of changing it again since display name != username. This problem could be alleviated by having users log in with their email but not entirely sure what exactly we want to do with usernames since they might just be duplicates of author display names. This would also mean having to do another POST request to update a user's username. Would prefer if there was a way to avoid this so we don't have duplicate data in our database.


front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+                            id="displayName"
+                            placeholder="New Display Name"
+                            onChange={handleInputChange}
+                            value={displayName}
+                        />
+                    </FormGroup>
+                    <FormGroup>
+                        <Label for="githubUrl">GitHub Account</Label>
+                        <Input
+                            type="text" // add form validation?
```
raysarinas: not sure if we want to add some form of form validation to ensure that a valid URL is submitted by a user

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+                }
+            });
+        }).catch(err => {
+            console.log("GET ERROR");
+        });
+
+        authorObject._id = adata._id;
+        authorObject.displayName = displayName;
+        authorObject.url = adata.url;
+        authorObject.github = githubUrl;
```
jmmxp: Could you pull from main and then try POST'ing an object without id, url, and type? I updated the endpoint to match our spec document yesterday and it should work with only taking in displayName+github.

front-end/src/pages/SettingsPage.tsx
jmmxp: I think we can leave this for later if we have some extra time. Maybe throw a TODO on it?

jmmxp (commented):
Will leave final review for Chris since he's more experienced in React.

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+    // https://reactstrap.github.io/components/form/#app
+    // https://medium.com/better-programming/easily-create-a-form-with-react-hooks-1cab17e2be0d
+    // setting state: https://stackoverflow.com/questions/45850550/accessing-data-from-axios-get-request
+
+    const initialInputState = { displayName: "", githubUrl: "" };
+    const [eachEntry, setEachEntry] = useState(initialInputState);
+    const { displayName, githubUrl } = eachEntry;
+    const authorObject = {
+        _id: "",
+        _type: "author",
```
VinnyLuu: same with @Jmmxp comment, dont need these fields anymore since the endpoint doesn't require them.

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+        setEachEntry({ ...eachEntry, [e.target.name]: e.target.value });
+    };
+
+    const handleSubmit = (e: any) => {
+        e.preventDefault();
+
+        axios.get(process.env.REACT_APP_API_URL + "/api/authors/").then(res => {
+            res.data.results.forEach((author: any) => {
+                // test by changing loggedInUser to a hardcoded string
+                if (author.displayName == loggedInUser.loggedInUser.username) {
```
VinnyLuu: The authors display name and the username of the user are completely different right? But this wont be needed anymore with @Jmmxp and @ChrisChrisLoLo change with the new endpoint since you'll have the author object

jmmxp: They're only the same right now as of the PR since it's set that way in signals.py.
It's a good point though, we need @ChrisChrisLoLo probably to change the registration form to also send github/url and make sure we properly set those in signals.py

But yes, we'll have the author object from the server soon, PR just went up

raysarinas:
Settings page can be found at `/settings`. Changeable fields are just in a Card
Screen Shot 2021-02-23 at 1 38 43 AM

Successful Post Request with both fields changed:
Screen Shot 2021-02-23 at 1 39 43 AM

Successful Post Request with only GitHub URL changed (Display Name input field is empty):
Screen Shot 2021-02-23 at 1 43 45 AM

Successful Post Request with only Display Name changed (GitHub URL is empty):
Screen Shot 2021-02-23 at 1 44 34 AM

ChrisChrisLoLo:
You bring up fair concerns in your PR description. We can talk on Discord about what the current approach is to access authors.

To be oversimple in my explanation, we now have the users author ID stored in the front-end. We can use the ID for everything as it's immutable unlike the display name. I don't think we need to iterate over all users on the settings page, since I think this should be for our users only (other servers will have a settings page for thier users.)

If you'd like to talk about it in higher latency, feel free to message the group chat or get in a voice call. I'm hoping the planned solution addresses most of your concerns, but I might not have addressed everything

front-end/src/components/AppNavBar.jsx
```diff
@@ -55,6 +55,9 @@ export default class NavbarComp extends React.Component {
           <NavbarToggler onClick={this.toggle}/>
           <Collapse isOpen={this.state.isOpen} navbar>
             <Nav className="ml-auto" navbar>
+            <NavItem onClick={this.close}>
```
ChrisChrisLoLo: Should this `NavItem` only appear when the user is logged in? If so, we can move this component to be beside the `Log Out` link

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+    Card,
+    CardBody,
+    Button,
+    Form,
+    FormGroup,
+    Input,
+    Label,
+} from 'reactstrap';
+
+const SettingsPage = (loggedInUser: any) => {
```
ChrisChrisLoLo: `loggedInUser` here can be `props`, since what we're passing in here is props

Bonus points if you can type the props parameter. An example of how to do this can be found here https://github.com/CMPUT404W21T02/group-project/blob/ba3760e021cff78cc8eb5944760e21d5885f8a3f/front-end/src/components/PostListItem.tsx#L6

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+        }).then(res => {
+            console.log(res);
+            console.log(res.data);
+        }).catch(err => {
+            console.log("POST ERROR");
+        });
+    };
+
+    return (
+        <Card>
```
ChrisChrisLoLo, line 77: Good use of Reactstrap!

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+const SettingsPage = (loggedInUser: any) => {
+    // https://reactstrap.github.io/components/form/#app
+    // https://medium.com/better-programming/easily-create-a-form-with-react-hooks-1cab17e2be0d
+    // setting state: https://stackoverflow.com/questions/45850550/accessing-data-from-axios-get-request
+
+    const initialInputState = { displayName: "", githubUrl: "" };
+    const authorUrl = process.env.REACT_APP_API_URL + "/api/author/" + loggedInUser.loggedInUser.authorId + "/";
+    const [eachEntry, setEachEntry] = useState(initialInputState);
+    const { displayName, githubUrl } = eachEntry;
+    const [unchangedData, setUnchangedData] = useState(initialInputState);
```
ChrisChrisLoLo, line 28: There are ways where we can achieve the same behaviour without `unchangedData` and `setUnchangedData`.

We can leave the code as is since it works, but if you'd curious about other ways to approach this pattern, feel free to message me

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const handleInputChange = (e: any) => {
+        setEachEntry({ ...eachEntry, [e.target.name]: e.target.value });
+    };
+
+    const handleSubmit = (e: any) => {
+        e.preventDefault();
+        getAuthorData();
+
+        axios.post(authorUrl, {
```
ChrisChrisLoLo: I think we should have some sort of UI to indicate if any error has occurred, or if the update was successful. This should make our UI functional without looking at any console logs

We could have some sort of `postError` state, and show a relevant message (it can even be a p tag for now, or better yet an alert (https://reactstrap.github.io/components/alerts/)) if sending an update failed

ChrisChrisLoLo (changes requested):
Good work for a first FE PR!

I liked how you used ReactStrap, and that you worked with the Hooks API

The two comments that I'll be a bit firm on is making the `settings` button visibile only to logged in users, as well as showing some UI indication if updating the profile was successful

I think the PR is good to be merged after this round

front-end/src/pages/SettingsPage.tsx
ChrisChrisLoLo: Thinking about it a bit I take it back. This isn't a pattern that I know the exact code for

front-end/src/pages/SettingsPage.tsx
```diff
@@ -0,0 +1,84 @@
+import axios from 'axios';
```
ChrisChrisLoLo, line 1: arbitrary comment to start a review

ChrisChrisLoLo (approved):
For the scope of the project I'll mark this as approve since this PR doesn't cause any regressions, and I want to try reducing overhead.

I still think you should consider adding the suggested changes (or push back if you'd like, and open a conversation about the feedback) before merging your PR in. We'll see how this style of review goes 😊

front-end/src/components/AppNavBar.jsx
raysarinas: done. `Log Out` is kinda weird now though (shows hovering over text, but the button still works)

Screen Shot 2021-02-24 at 1 39 24 AM

front-end/src/pages/SettingsPage.tsx
raysarinas: done. only thing is that now the settings page can be accessed by someone who isn't logged in (via `localhost:8000/settings`) but if they try to do anything, an error will always be displayed/alerted

Screen Shot 2021-02-24 at 2 03 21 AM

front-end/src/pages/SettingsPage.tsx
raysarinas: done:

On success:
Screen Shot 2021-02-24 at 2 04 08 AM

On empty input:
Screen Shot 2021-02-24 at 2 04 35 AM

On error (i.e. not logged in, but trying to access settings page, or if an axios request fails that sort of thing):
Screen Shot 2021-02-24 at 2 03 21 AM

front-end/src/components/AppNavBar.jsx
ChrisChrisLoLo: Sweet! The logout part is kind of weird (that's on me, I need to figure out how to get it to work with Reactstrap), so I can fix it up at a later point

front-end/src/pages/SettingsPage.tsx
ChrisChrisLoLo: Looks good!

We can create something in React router that runs a conditional and does a redirect accordingly
https://reactrouter.com/web/example/auth-workflow.

I'm personally okay if we don't account for this edgecase, but if you'd like to tackle it at some point the link above should give some clues about where to start

front-end/src/pages/SettingsPage.tsx
ChrisChrisLoLo: Great!

PR #59, opened 2021-02-21
Change Author ID to UUID field

jmmxp:
Changed Author ID to UUID. From Hindle's responses, it doesn't seem like there will be any use-case where we need to be storing this as an actual whole URL (since our server only stores Authors created on our own client).


VinnyLuu (approved):
LGTM

PR #65, opened 2021-02-22
Change ID to UUID field for Post model

jmmxp:
Changed ID to UUID field for Post model and fixed up a couple of other fields that shouldn't have had `unique=True`.

Will PR in the future to fix author to foreign key. We might have to hash out how shared posts work - do we take the author from the shared post object?


back-end/api/quickstart/models.py
```diff
@@ -24,12 +24,12 @@ class Visibility(models.TextChoices):
...
-  title = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  title = models.TextField()
   description = models.TextField()
-  source = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
-  origin = models.CharField(max_length=LONG_CHAR_LENGTH, unique=True)
+  source = models.TextField()
+  origin = models.TextField()
   visibility = models.CharField(max_length=SHORT_CHAR_LENGTH, choices=Visibility.choices)
   unlisted = models.BooleanField()
   author = models.CharField(max_length=LONG_CHAR_LENGTH)
```
VinnyLuu, line 35: should these other char fields be changed to textfields as well?

jmmxp: Gonna leave this for a future PR for someone to change all of them

Didn't wanna pollute this PR (I changed these ones on this PR since they shouldn't have been `unique=True`)

VinnyLuu (approved):

PR #68, opened 2021-02-22
Fixed Create Post on the front end

VinnyLuu:
Fixed the create post component to properly hit the /api/author/authorid/posts/ endpoint to POST a post, as well as fixed the component itself. 

Accessible through the homepage with a button
image

Create new post
image

### DIFF
--- a/front-end/src/components/CreatePost.tsx
+++ b/front-end/src/components/CreatePost.tsx
@@ -63,9 +63,11 @@ const CreatePostComponent = (props: any) => {
                 <Input type="text" name="Description" placeholder="Description" onChange={e => setDesc(e.target.value)}/>
                 <Label for="Content Type">Content Type</Label>
                 <select name="Content Type" onChange={e => setContentType(e.target.value)}>
-                    <option value="Text">Text</option>
-                    <option value="Markdown">Markdown</option>
-                    <option value="Image">Image</option>
+                    <option value="text/markdown">text/markdown</option>
+                    <option value="text/plain">text/plain</option>
+                    <option value="application/base64">application/base64</option>
+                    <option value="image/png;base64">image/png</option>
+                    <option value="image/jpeg;base64">image/jpeg</option>
                 </select>
                 <Input type="text" name="Content" placeholder="Content" onChange={e => setContent(e.target.value)}/>
                 <Input type="text" name="Categories" placeholder="Categories" onChange={e => parseCategories(e.target.value)}/>
--- a/front-end/src/components/CreatePost.tsx
+++ b/front-end/src/components/CreatePost.tsx
@@ -1,6 +1,6 @@
 import axios from 'axios';
 import React, { useState } from 'react';
-import { Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
+import { Alert, Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
 import { Post } from '../types/Post';
 import { Form, Input } from 'reactstrap';
 
@@ -8,11 +8,12 @@ const CreatePostComponent = (props: any) => {
 
     const [title, setTitle] = useState('');
     const [desc, setDesc] = useState('');
-    const [contentType, setContentType] = useState('');
+    const [contentType, setContentType] = useState('Text');
     const [content, setContent] = useState('');
-    const [categories, setCategories] = useState('');
+    const [categories, setCategories] = useState<string[]>([""])
     const [visibility, setVisibility] = useState('Public');
     const [unlisted, setUnlisted] = useState(false);
+    const [showError, setShowError] = useState(false);
 
     function changeVisibility(isChecked: boolean) {
         if (isChecked) {
@@ -23,7 +24,14 @@ const CreatePostComponent = (props: any) => {
         }
     }
 
+    function parseCategories(categories: string) {
+        let parsed: string[] = categories.split(" ")
+        setCategories(parsed);
+    }
+
     function sendPost(e:any) {
+        e.preventDefault();
+
         let data = {
             title: title,
             description: desc,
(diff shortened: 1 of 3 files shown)


front-end/src/components/CreatePost.tsx
```diff
@@ -4,26 +4,37 @@ import { Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
...
     const [desc, setDesc] = useState('');
     const [contentType, setContentType] = useState('');
     const [content, setContent] = useState('');
     const [categories, setCategories] = useState('');
-    const [visibility, setVisibility] = useState('');
-    const [unlisted, setUnlisted] = useState('');
+    const [visibility, setVisibility] = useState('Public');
+    const [unlisted, setUnlisted] = useState(false);
+
+    function changeVisibility(isChecked: boolean) {
```
ChrisChrisLoLo, line 18: Good use of the checkbox!

front-end/src/components/CreatePost.tsx
```diff
@@ -4,26 +4,37 @@ import { Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
...
             visibility: visibility,
             unlisted: unlisted,
             contentType: contentType,
             content: content,
-            categories: categories
-        })
+            categories: [categories]
+        }
+        
+        axios.post(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/", data)
```
ChrisChrisLoLo, line 45: We could have some sort of basic UI in the case that a post fails for some reason, kind of like this:  https://github.com/CMPUT404W21T02/group-project/blob/ba3760e021cff78cc8eb5944760e21d5885f8a3f/front-end/src/components/LoginForm.jsx#L63.

What are your thoughts?

front-end/src/pages/HomePage.tsx
```diff
@@ -57,6 +59,9 @@ export default function HomePage(props:RouteComponentProps) {
         <h3>Feed</h3>
         {postListElToDisplay}
       </Col>
+      <Col>
+        <Button onClick={() => createPost("/create_post")}>Create Post</Button>
```
ChrisChrisLoLo: Nit: could we replace `createPost()` with `props.history.push('/create_post');`?

front-end/src/components/CreatePost.tsx
```diff
@@ -4,26 +4,37 @@ import { Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
...
+    }
 
     function sendPost(e:any) {
-        axios.post(process.env.REACT_APP_API_URL + "/api/author/username/posts/", {
+        let data = {
             title: title,
             description: desc,
             visibility: visibility,
             unlisted: unlisted,
             contentType: contentType,
```
ChrisChrisLoLo, line 40: It might be useful to have contentType as a drop-down, as it reduces potential user error

We could do this at a later point if you'd like (after this PR), but I think it could potentially make demoing go a bit smoother

front-end/src/components/CreatePost.tsx
```diff
@@ -4,26 +4,37 @@ import { Label, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
...
             content: content,
-            categories: categories
-        })
+            categories: [categories]
```
ChrisChrisLoLo: We might need to create a new issue about how to go about this. Currently if I type `'images, pets'` into the categories field, it will be sent to the server as 
```javascript
 ['images, pets']
``` 
rather than 
```javascript
['images','pets']
```
That being said categories doesn't seem to be a big deal in the user stories. What do you think about this?

ChrisChrisLoLo (approved):
Looks good!

I'll left a few comments. The one that I'm most firm about is having some sort of basic UI to notify you of the status of your post. The rest I'm more flexible about

Feel free to merge once you've reviewed the comments and made changes to the ones you want to change

front-end/src/components/CreatePost.tsx
VinnyLuu: Yup sounds good, I've added an Alert similarly

front-end/src/pages/HomePage.tsx
VinnyLuu: Done

front-end/src/components/CreatePost.tsx
VinnyLuu: Sounds good, I've changed the contentType to a drop-down :)

front-end/src/components/CreatePost.tsx
VinnyLuu: I've changed this up for now to be split by whitespace. I think for now its okay, but when this gets changed to a modal this can be more fully fleshed out. I was thinking for this it could use something like adding tags: https://www.npmjs.com/package/react-tag-input

ChrisChrisLoLo: Good idea! As you've said, that would be pretty nice to have later on

front-end/src/components/CreatePost.tsx
```diff
@@ -23,29 +24,51 @@ const CreatePostComponent = (props: any) => {
...
         <div>
+            {showError ? <Alert>Could not create post</Alert> : null}
             <h1>Create Post</h1>
             <Form inline={true} onSubmit={e => sendPost(e)}>
                 <Input type="text" name="Title" placeholder="Title" onChange={e => setTitle(e.target.value)}/>
                 <Input type="text" name="Description" placeholder="Description" onChange={e => setDesc(e.target.value)}/>
-                <Input type="text" name="Content Type" placeholder="Content Type" onChange={e => setContentType(e.target.value)}/>
+                <Label for="Content Type">Content Type</Label>
+                <select name="Content Type" onChange={e => setContentType(e.target.value)}>
+                    <option value="Text">Text</option>
```
ChrisChrisLoLo: Dropdown looks good!
I could have better clarified this, but according to one of Justins comments on the spec document
```
from the Post JSON defined in the spec, only the following are possible:

# text/markdown -- common mark
# text/plain -- UTF-8
# application/base64
# image/png;base64 # this is an embedded png -- images are POSTS. So you might have a user make 2 posts if a post includes an image!
# image/jpeg;base64 # this is an embedded jpeg
```
Could you make the values for the dropdown "image/jpeg;base64" and such? If need be we can change this later when porting it to a modal

front-end/src/pages/HomePage.tsx
```diff
@@ -60,7 +57,7 @@ export default function HomePage(props:RouteComponentProps) {
         {postListElToDisplay}
       </Col>
       <Col>
-        <Button onClick={() => createPost("/create_post")}>Create Post</Button>
+        {props.loggedInUser ? <Button onClick={() => props.history.push('/create_post')}>Create Post</Button> : null}
```
ChrisChrisLoLo, line 60: Good catch!

ChrisChrisLoLo (approved):
Looks great!

My only comment would be to change the values of the "Content-Type" dropdown. This can be changed later if you'd like, so feel free to merge it in when you're ready.

PR #69, opened 2021-02-24
Change Comment ID field to UUID

jmmxp:
No need to use `CharField` for this as we can be guaranteed that comments stored in our database are for posts on our server. When we connect with other teams though, we might have to change the serializer to have the JSON response have the full URL+ID though, instead of just this UUID.


back-end/api/quickstart/models.py
```diff
@@ -42,20 +42,14 @@ class Visibility(models.TextChoices):
   commentLink = models.TextField(default='')
 
 class Comment(models.Model):
-  id = models.CharField(primary_key=True, max_length=LONG_CHAR_LENGTH)
+  id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
raysarinas, line 45: could we use `uuid.uuid4.hex` instead as default? the `.hex` removes all the arbitrary dashes in the UUID. but up to you; doesn't really matter

raysarinas (approved):
LGTM, pls see comment though

back-end/api/quickstart/models.py
jmmxp: https://stackoverflow.com/questions/48438569/returning-hex-uuid-as-default-value-for-django-model-charfield 
leaving this for reference if we want to change to hex, I'll leave as is for now since using hex actually makes the tests start failing

PR #70, opened 2021-02-24
Move endpoint tests to directory tests/endpoints

jmmxp:
For better organization of our code. Also planning to add tests for models, so having this `tests` directory this is a precursor to that.


raysarinas (approved):
👍🏼

PR #78, opened 2021-02-25
Add tests for all models

jmmxp:
Added model tests that test creation of each model we currently have, and ensures the created object's fields are as expected.

### DIFF
--- a/back-end/api/quickstart/tests/helper_test.py
+++ b/back-end/api/quickstart/tests/helper_test.py
@@ -4,7 +4,8 @@ def get_test_author_fields(i=1):
   return {
     'displayName': f'John Doe {i}',
     'url': f'testUrl{i}',
-    'github': f'testGithub{i}'
+    'github': f'testGithub{i}',
+    'host': 'testHost'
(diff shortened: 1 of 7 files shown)


back-end/api/quickstart/tests/helper_test.py
```diff
@@ -55,3 +56,9 @@ def get_test_partial_post_fields(i=1, author="testauthorid"):
     'categories': '["Testing"]',
     'commentLink': 'link to comments'
   }
+
+def get_test_inbox_fields():
+  return {
+    'author': 'testAuthor',
+    'items': '[{"type":"post"}, {"type":"comment"}]'
```
VinnyLuu, line 63: missing likes?

VinnyLuu (approved):
LGTM

back-end/api/quickstart/tests/helper_test.py
jmmxp: since it's a JSON field it can take in anything, so no need to have likes here

Issue #61, opened 2021-02-21
Use TextField Instead of CharFields on Backend Models


raysarinas:
Make sure to have form validation on the frontend.

PR #82, opened 2021-02-26
Migrate models' CharFields to TextFields

jmmxp:
Any CharField length limit we have is completely arbitrary, indicating that we should just use TextFields instead. We need to make sure we sanity check / enforce limits client-side eventually though.


raysarinas:
Closes #61

VinnyLuu (approved):
LGTM

PR #83, opened 2021-02-26
Change post author to ForeignKey, update tests accordingly

jmmxp:
Since we're guaranteed that posts on our server must always be created by an author on our server, we can make a post's author into a ForeignKey field. We do this because, according to Dr. Hindle, data for posts on our server should always be up to date. (i.e., the author data of that post should also be up to date).

Note that posts in an author's inbox could be from other servers, but we are assuming in that case that those posts sent to the inbox from other servers have proper JSON and we simply store those posts exactly as they are given to us.


VinnyLuu (approved):
LGTM

PR #84, opened 2021-02-27
Update get posts endpoint and add basic authentication

### LINKED ISSUE #79: Endpoint to view posts of author X as author Y
### Related User Stories
**37. As an author, I should be able to browse the public posts of everyone.**

jmmxp:
When getting posts for an author, if the queried author is the same as the authenticated/logged in author, we get *all* posts for that author. Otherwise we only get public, unlisted=False posts. 

I've added basic authentication to the whole viewset (so creation of posts requires authentication now as well).

Closes #79.

### DIFF
--- a/back-end/api/quickstart/views.py
+++ b/back-end/api/quickstart/views.py
@@ -1,11 +1,12 @@
 from django.contrib.auth.models import User, Group
 from .models import Author, Post, Follow, Comment, Like, Inbox
 from rest_framework import viewsets
-from rest_framework import permissions
 from rest_framework import status
 from rest_framework.response import Response
 from quickstart.serializers import UserSerializer, GroupSerializer, AuthorSerializer, PostSerializer, FollowSerializer, CommentSerializer, LikeSerializer, InboxSerializer
 from .mixins import MultipleFieldLookupMixin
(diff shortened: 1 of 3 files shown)


back-end/api/quickstart/views.py
```diff
@@ -1,11 +1,12 @@
 from django.contrib.auth.models import User, Group
 from .models import Author, Post, Follow, Comment, Like, Inbox
 from rest_framework import viewsets
-from rest_framework import permissions
 from rest_framework import status
 from rest_framework.response import Response
 from quickstart.serializers import UserSerializer, GroupSerializer, AuthorSerializer, PostSerializer, FollowSerializer, CommentSerializer, LikeSerializer, InboxSerializer
 from .mixins import MultipleFieldLookupMixin
+from rest_framework.authentication import SessionAuthentication, BasicAuthentication
```
VinnyLuu: SessionAuthentication unused?

VinnyLuu (approved):
LGTM

back-end/api/quickstart/views.py
jmmxp: Good catch, removed.

Issue #81, opened 2021-02-26
Hide Unlisted Posts


raysarinas:
### Related User Stories
**43. As an author, I want to be able to make posts that are unlisted, that are publicly shareable by URI alone (or for embedding images)**

PR #85, opened 2021-02-27
Change public post to not show unlisted post

VinnyLuu:
The public post and post list endpoint now doesn't show unlisted posts. Updated test accordingly. Closes (#81 )

### DIFF
--- a/back-end/api/quickstart/tests/endpoints/test_public_post_list.py
+++ b/back-end/api/quickstart/tests/endpoints/test_public_post_list.py
@@ -14,22 +14,19 @@ class GetPublicPosts(TestCase):
     self.author2 = Author.objects.create(**get_test_author_fields())
 
     self.post1 = Post.objects.create(
-      **get_test_post_fields(1, visibility="Public"), author=self.author1
+      **get_test_post_fields(1, visibility="Public", unlisted=False), author=self.author1
     )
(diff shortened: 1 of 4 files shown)


back-end/api/quickstart/tests/endpoints/test_public_post_list.py
```diff
@@ -25,15 +25,21 @@ def setUp(self):
     self.post4 = Post.objects.create(
       **get_test_post_fields(4, visibility="Private"), author=self.author2
     )
```
jmmxp: nit; we probably should have changed this before but maybe set unlisted=False on all these previous posts to help make it more explicit which post should be returned in the test.

As it is right now, we ensure the response has 3 posts returned but it's not exactly clear which ones are gonna be returned esp. cause these posts don't have unlisted explicitly set (and we wouldn't know if the default is unlisted=False/True) without having to check

jmmxp (approved):
LGTM just a nit

back-end/api/quickstart/tests/endpoints/test_public_post_list.py
VinnyLuu: Thanks, done. It showed that the new post I added was the same as the first one, which was redundant so I removed it.

Issue #50, opened 2021-02-19
Refactor JSX Files to TSX


raysarinas:
We have a handful of JSX files that should be refactored to TSX. 

Current JSX files don't use react hooks as far as I can tell and still use "older" conventions of using classes.

Refactoring these files will probably be good practice for writing react in typescript. 

Related Files:

- ~~front-end/src/components/AppNavBar.jsx~~
- front-end/src/components/LoginForm.jsx
- front-end/src/components/RegisterForm.jsx
- front-end/src/pages/AuthPage.jsx

PR #87, opened 2021-02-28
Refactor Navbar

raysarinas:
There was a weird thing going on with the `Log Out` button and I figured I might as well refactor the JSX file to TSX. Didn't take very long.

There is an issue with the navbar links showing up vertically stacked instead of horizontally in a row. Not sure what the problem is because this problem persists even before the refactoring. We can figure it out later when we tidy up the UI. 

Navbar should look exactly the same.

Related to Issue #50


ChrisChrisLoLo (approved):
LGTM!

Nice use of the `Props` typing!

PR #88, opened 2021-02-28
Add Basic Auth to Making Posts

raysarinas:
This will fix any `401` response errors. If it doesn't, then it's because we need to add headers


raysarinas:
Also fixed a small bug that would cause react to crash or be weird when trying to access the `create_post` page

jmmxp (approved):
LGTM

PR #90, opened 2021-02-28
Create edit/create modal

ChrisChrisLoLo:
Add functionality to the create modal to allow it to make edits to posts.

Note that editing posts have not yet been tried out yet, and may need small adjustments for it to work.

Some parts of the modal have already been merged

Can now upload images via file upload.

Added tweaks to `PostDetailModal` to keep elements within the modal

Screenshot_20210228_110928
Screenshot_20210228_110738

### DIFF
--- a/front-end/src/components/CreateEditPostModal.tsx
+++ b/front-end/src/components/CreateEditPostModal.tsx
@@ -107,7 +107,7 @@ export default function CreateEditPostModal(props: Props){
           })
       }
       else if(props.editFields !== undefined){
-          axios.put(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/" + props.editFields.id + "/", data, config)
+          axios.post(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/" + props.editFields.id + "/", data, config)
           .then(res => {
             handleRes(res)
           }).catch(error => {
@@ -118,16 +118,19 @@ export default function CreateEditPostModal(props: Props){
   }
 
   function handleRes(res:AxiosResponse){
-    if (res.status >= 400) {
+    if (res.status >= 300) {
       setShowError(true)
-    } else if (res.status === 201) {
+    } else if (res.status >= 200) {
       if(isCreate){
         const post:Post = res.data;
         if(props.prependToFeed !== undefined){
           props.prependToFeed(post)
         }
+        resetFormAndToggle()
+      } else {
+        // We do not want to wipe the fields on our form when we edit
+        props.toggle()
       }
-      resetFormAndToggle()
     }
   }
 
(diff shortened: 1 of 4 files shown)


front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -1,23 +1,47 @@
...
+  const isCreate = props.editFields === undefined
+
+  const [title, setTitle] = useState(postFields.title);
+  const [desc, setDesc] = useState(postFields.description);
+  const [contentType, setContentType] = useState(postFields.contentType);
+  const [content, setContent] = useState(postFields.content);
+  const [categories, setCategories] = useState<string[]>(postFields.categories);
+  const [visibility, setVisibility] = useState(postFields.visibility);
+  const [unlisted, setUnlisted] = useState(postFields.unlisted);
   const [showError, setShowError] = useState(false);
```
raysarinas, line 45: I think we also need to be able to have author's be able to create a post with a given post_id (in the spec, we should PUT to create a post with an author specified ID. I'm not sure if this will require a change on the backend model.

front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -53,52 +98,89 @@ export default function CreatePostComponent(props: Props){
...
+      if(isCreate){
+        axios.post(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/", data, config)
+          .then(res => {
+            handleRes(res)
+          }).catch(error => {
+            setShowError(true)
+          })
+      }
+      else if(props.editFields !== undefined){
+          axios.put(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/" + props.editFields.id + "/", data, config)
```
raysarinas: editing posts should do a POST request and PUT should be only used when a author-specified post ID is present

front-end/src/pages/HomePage.tsx
```diff
@@ -37,6 +37,12 @@ export default function HomePage(props:any) {
     props.history.push('/authors/'+userSearch);
   }
 
+  function preprendToFeed(post:Post){
```
raysarinas: typo? did you mean prepend instead of preprend? lol

front-end/src/components/CreateEditPostModal.tsx
raysarinas: on second thought - we can add author-specified post_id's later. nbd, I'll just open up a new/separate issue for it since I think it can be worked on independently from everything else in this PR/issue

raysarinas (approved):
LGTM, just take a look at comments I guess and we can discuss or address the concerns later

front-end/src/components/CreateEditPostModal.tsx
ChrisChrisLoLo: Good catch! I agree that a separate issue would be good. Hopefully, it should be small enough for anyone to do if they have some time

front-end/src/components/CreateEditPostModal.tsx
ChrisChrisLoLo: This might explain the server issue that I was having. Let me play around with it for a bit

PR #91, opened 2021-02-28
Display commonmark content

ChrisChrisLoLo:
Depends on #90 

Displays the posts in Commonmark

- Light refactor to helper functions, makes use of TS enums
- Add commonmark functionality
- Reverses the order of obtain posts

Screenshot_20210228_155850


raysarinas (approved):
LGTM! Like the use of enums instead of hardcoded strings 💯

PR #86, opened 2021-02-28
View author profile page

raysarinas:
Related to #46 

Clicking the `Profile` button redirects the user to `/author/{AUTHOR_ID}/` and the following gets displayed. Screenshot shows hovering over the GitHub URL, which you can see in the bottom left actually will redirect whoever is clicking the link to that URL: 
Screen Shot 2021-02-27 at 7 24 49 PM

- WASSUP == displayname
- @ tester123 == username

If some error occurs when getting the author's info, the only thing that is displayed is an error message:
Screen Shot 2021-02-27 at 7 33 56 PM

If someone who isn't logged in tries to access the user's profile, they get a 404. Not sure if we want profiles to be public:
Screen Shot 2021-02-27 at 7 25 33 PM

PR #101, opened 2021-03-03
Remove old nav bar

jmmxp:
Old JSX navbar was accidentally added back in #86, so we have 2 NavBar files now. I tested the site after removing and everything seems to be fine.


raysarinas:
Yeah, I tested everything when adding the new nav bar. Should have no problems with it, probably just something that confused git during a PR merge

raysarinas (commented):
LGTM

PR #102, opened 2021-03-03
Implemented deleting post UI

### LINKED ISSUE #72: UI for Deleting Posts
### Related User Stories
**17. As an author, I want to delete my own public posts.
24. As an author, I want t be able to use my web-browser to manage my posts.**

### Description / Notes
There might be a number of ways to do this, like having a dropdown with the option to edit/delete, or a lil trash can or X button that deletes a post. 

- DELETE ://service/author/{AUTHOR_ID}/posts/{POST_ID} - remove the post
- Delete posts from frontend via button or dropdown button
- Alert/Notify user on error; i.e., if unable to delete or delete request didn't go through
- Test UI for deleting public posts

### Tasks
- [ ] Be able to delete posts from the frontend
- [ ] Modify and re-run tests if necessary

VinnyLuu:
Users can now delete posts from both the homepage and the author page. Refactored PostList into a new component, since both the author page and the home page are using them. Deleting a post will automatically update the list as well.

Delete modal
image

Error msg
image

### DIFF
--- a/front-end/src/components/PostList.tsx
+++ b/front-end/src/components/PostList.tsx
@@ -1,0 +1,62 @@
+import React, { useState } from 'react';
+import { Row, Col, Button } from 'reactstrap';
+import CreateEditPostModal from '../components/CreateEditPostModal';
+import PostListItem from '../components/PostListItem';
+import { Post } from '../types/Post';
+
+
+export default function PostList(props:any) {
+
+  const [isCreatePostModalOpen, setIsCreatePostModalOpen] = useState<boolean>(false);
+
+  function prependToFeed(post:Post){
+    if(props.postEntries !== undefined){
+      props.setPostEntries([post, ...props.postEntries])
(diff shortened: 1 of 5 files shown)


jmmxp:
Can you make the error message red?

front-end/src/components/PostList.tsx
```diff
@@ -0,0 +1,75 @@
...
+  postEntries?: Post[];
+  setPostEntries?: Function;
+  loggedInUser?: UserLogin;
+}
+
+/**
+ * Post list component to show list of posts
+ * @param props 
+ */
+export default function PostList(props:any) {
```
jmmxp: props: Props

jmmxp (changes requested):

jmmxp (approved):
LGTM you can edit PR desc. now that we found a way to revert the UI changes

PR #105, opened 2021-03-03
Search authors

Monica-Uwalaka:
Can now search for multiple authors by displayName. Clicking on an author takes to to a page to view their details. For now, I just have the id on there.
Screenshot (133)
Screenshot (132)


front-end/public/index.html
```diff
@@ -29,6 +29,7 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
+    <div id = "belowRoot"></div>
```
raysarinas: Was this manually added? What is this for?

front-end/src/App.scss
```diff
@@ -1 +1,5 @@
 @import "~bootstrap/scss/bootstrap";
+@import '~antd/dist/antd.css';
+@import './assets/scss/main.scss';
```
raysarinas: where are these being used? Using Ant Design together with Bootstrap/Reactstrap might cause our UI to break or act weird but I'm not sure. Chris probably knows

front-end/src/App.tsx
```diff
@@ -12,6 +12,7 @@ import { UserLogin } from './types/UserLogin';
 import AuthorPage from './pages/AuthorPage';
 import SettingsPage from './pages/SettingsPage';
 import CreatePostComponent from './components/CreatePost';
+import AuthorDetail from './pages/AuthorDetail';
```
raysarinas: I don't think you added this page to git.

front-end/src/App.tsx
```diff
@@ -36,6 +37,7 @@ function App() {
               {loggedInUser && <Route path={`/author/${loggedInUser.authorId}`} render={(props) => <AuthorPage {...props} loggedInUser={loggedInUser} />} />}
               {/* TODO: hide settings page if not logged in */}
               <Route path="/settings" render={(props) => <SettingsPage loggedInUser={loggedInUser} />} />
+              <Route path="/author_detail/:id" render={(props) => <AuthorDetail {...props}/>} />
```
raysarinas: The route should just be `"/author/:id/"`, which already exists. Right now we should be able to search ourselves up (i.e., if you are logged in, you should be able to search for yourself and click on your page) but it will just show a "404" if you try to view any other author

front-end/src/pages/HomePage.tsx
```diff
@@ -1,17 +1,23 @@
 import axios from 'axios';
 import React, { useEffect, useState } from 'react';
+import ReactDOM from "react-dom"
 import { RouteComponentProps } from 'react-router-dom';
 import { Row, Col, Form, Input, Button } from 'reactstrap';
 import CreateEditPostModal from '../components/CreateEditPostModal';
 import PostListItem from '../components/PostListItem';
 import { Post } from '../types/Post';
-
+import AuthorList from "../components/AuthorList"
```
raysarinas: I don't think AuthorList was added to git either

front-end/src/pages/HomePage.tsx
```diff
@@ -1,17 +1,23 @@
...
+interface userProps {
+  url: string;
+  displayName: string;
+  id: string;
+  github: string;
+}
```
raysarinas: Please use the `Author` type (`import Author from "../components/Author"`)

You can see how to use it inside `AuthorPage.tsx`

front-end/src/pages/HomePage.tsx
```diff
@@ -32,9 +38,12 @@ export default function HomePage(props:any) {
     setUserSearch(e.target.value);
   }
 
+  
+  //search for authors with the same displayName
   function searchUsers(e:any){
     e.preventDefault();
-    props.history.push('/authors/'+userSearch);
+    const URL = process.env.REACT_APP_API_URL + "/api/authors/";
+    ReactDOM.render(<AuthorList url = {{searchURL: URL}}/>, document.getElementById("belowRoot"));
```
raysarinas: Not sure if we should be doing this and I don't know what would be better to do - maybe someone else can help. I think we should be navigating to the `AuthorPage`?

front-end/src/pages/HomePage.tsx
```diff
@@ -63,6 +72,37 @@ export default function HomePage(props:any) {
   }
   console.log(postListElToDisplay)
 
+  
+  function result(){
+    if (userSearch){
+      return (
+        <>
```
raysarinas: should this be `<div>` instead of `<>`?

raysarinas (changes requested):
Please address comments.

front-end/src/App.scss
ChrisChrisLoLo: Ideally we stick to one CSS framework. Bootstrap should be able to do what ant does. Here's an example of the Grid system in Bootstrap: https://getbootstrap.com/docs/4.0/layout/grid/. Let me know if you have any questions about mapping over Ant  concepts to Bootstrap

front-end/public/index.html
ChrisChrisLoLo: Feel free to put anything like footers into the App.tsx file

Monica-Uwalaka: @raysarinas yes it was manually added. I added this because I did not want to cover the search bar when showing the result of the search. @raysarinas @ChrisChrisLoLo is this alright? or is there a better way to do it?

front-end/src/App.scss
Monica-Uwalaka: Thanks for the link. I'll stick with Boostrap going forward.

front-end/src/pages/HomePage.tsx
Monica-Uwalaka: This url is used to get a list of authors with the same display name. The AuthorPage seems to only be accessible by a loggedInUser.

front-end/package.json
```diff
@@ -11,6 +11,7 @@
     "@types/react": "^16.9.53",
     "@types/react-dom": "^16.9.8",
     "@types/react-router-dom": "^5.1.7",
+    "antd": "^4.13.0",
```
ChrisChrisLoLo: Could you remove this dependency?

front-end/src/pages/HomePage.tsx
ChrisChrisLoLo: I think we should create a new page for the `authorList` component, and call it `AuthorSearchPage` or something similar. We can then render what you have here in that new page. With this new feature, we should be able to access the profile page without a loggedInUser.

Hopefully we already added logic to the `AuthorPage` if the `loggedInUser` is undefined

front-end/public/index.html
ChrisChrisLoLo: I think the best way to render the result is to put it into a Page component and put the Page into our router in `App.tsx`. It might be useful to cross-compare the other React pages we currently have to see how this is done

Monica-Uwalaka: I'll do that.

front-end/package.json
Monica-Uwalaka: yes

front-end/src/pages/HomePage.tsx
raysarinas: I don't think we've added that logic yet. I think for Friday it might be a good idea to just have search results displayed on a new page. Right now, we should be able to see a GET request to Django and then get an unauthorized response when trying to access another Author's page that isn't your own or if you aren't logged in. I don't think this response shows up on the frontend but it will on the Django console. Might need to double check this

front-end/src/pages/HomePage.tsx
```diff
@@ -1,11 +1,8 @@
 import axios from 'axios';
 import React, { useEffect, useState } from 'react';
-import ReactDOM from "react-dom"
-import { RouteComponentProps } from 'react-router-dom';
-import { Row, Col, Form, Input, Button } from 'reactstrap';
+import { Row, Col, Form, Input } from 'reactstrap';
 import PostList from '../components/PostList';
 import { Post } from '../types/Post';
-import AuthorList from "../components/AuthorList"
 interface userProps {
```
raysarinas: can this be changed to just have an `Author` object? I can do it. It's a quick fix

raysarinas (approved):
I think all good for now, just one quick comment

ChrisChrisLoLo:
@raysarinas Good catch! I refactored it accordingly

PR #106, opened 2021-03-04
Content Preview + Enhancements

ChrisChrisLoLo:
Create content preview for posts
- Change around typings
- Create content preview element
- Integrate element into the list view, detail view, and edit/create modal
- Fix navbar styling

NOTE: It may be easier to go by commit rather than by file when reviewing

#99
Screenshot_20210303_173758
Screenshot_20210303_173818
Screenshot_20210303_173929

### DIFF
--- a/front-end/src/components/PostListItem.tsx
+++ b/front-end/src/components/PostListItem.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { Button, Card, CardBody, CardLink, CardSubtitle, CardText, CardTitle } from 'reactstrap';
 import { Post } from '../types/Post';
 import { UserLogin } from '../types/UserLogin';
+import ContentPreview from './PostContentEl';
 import CreateEditPostModal from './CreateEditPostModal';
 import DeletePostModal from './DeleteModal';
 import PostDetailModal from './PostDetailModal';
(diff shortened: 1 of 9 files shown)


front-end/src/components/PostListItem.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { Button, Card, CardBody, CardLink, CardSubtitle, CardText, CardTitle } from 'reactstrap';
 import { Post } from '../types/Post';
 import { UserLogin } from '../types/UserLogin';
+import ContentPreview from './PostContentEl';
```
VinnyLuu: nit. In different files the naming is different, like in PostDetailModal its called PostContent, maybe just change the file and component name to just be ContentPreview or something and keep it consistent in the imports through all the files. I'm also not sure what the El at the end stands for.

VinnyLuu (approved):
LGTM just a nit, you can merge it in whenever you like :)

front-end/src/components/PostListItem.tsx
ChrisChrisLoLo: Good catch! I was renaming the component a lot, and that slipped through the cracks. El at the end stands for `Element` to avoid it being confused with the `PostContent` type. I'll see if I can have `PostContent` refer to both a type and component

VinnyLuu: Awesome sounds good. Great stuff on the preview. Looks good 👍

ChrisChrisLoLo: After trying it out, looks like there's a naming conflict if both the component and type is named `PostContent`. I'll keep it as `PostContentEl` for now, until we find a better name. IMO `El` is an unusual suffix

PR #107, opened 2021-03-04
Posts are now only editable/deletable if they are the author

VinnyLuu:
When choosing to display the edit/delete button, the code now checks if the loggedInUser.authorid matches with the posts authorid. Also removed the isEditable and isDeletable props.

### DIFF
--- a/front-end/src/components/PostListItem.tsx
+++ b/front-end/src/components/PostListItem.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Button, Card, CardBody, CardLink, CardSubtitle, CardText, CardTitle } from 'reactstrap';
+import { Card, CardBody, CardLink, CardSubtitle, CardText, CardTitle } from 'reactstrap';
 import { Post } from '../types/Post';
 import { UserLogin } from '../types/UserLogin';
 import CreateEditPostModal from './CreateEditPostModal';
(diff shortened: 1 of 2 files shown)


front-end/src/components/PostListItem.tsx
```diff
@@ -8,16 +8,14 @@ import PostDetailModal from './PostDetailModal';
...
-  isEditable?: boolean;
-  // true if the user can delete the post
-  isDeletable?: boolean;
   loggedInUser?: UserLogin;
   removeFromFeed: Function;
 }
 
 export default function PostListItem(props:Props) {
 
+  const isAuthorPost: boolean = (props.loggedInUser ? props.loggedInUser.authorId : "") === props.post.author.id;
```
ChrisChrisLoLo, line 17: Here's another way you can write this line using optional chaining and nullish coalescing.
I _think_ in the years to come this suggested way will be the preferred way to write it, but for now, it's probably non-standard and confusing.
This is more of a fun fact than a firm suggestion. No worries about writing it this way
```suggestion
  const isAuthorPost: boolean = props.loggedInUser?.authorId ?? "" === props.post.author.id;
```
NOTE: untested code

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

ChrisChrisLoLo (approved):
LGTM! I'll merge this as is! Nice refactoring!

PR #111, opened 2021-03-05
Add Comments to front-end

raysarinas:


front-end/src/components/PostDetailModal.tsx
```diff
@@ -9,6 +9,10 @@ interface Props {
   isOpen: boolean;
 }
 
+/**
+ * Post Details Modal to show the details of an individual post on a stream
+ * @param props 
+ */
 export default function PostListItem(props:Props) {
```
jmmxp: I think you're right, this should just be renamed to PostDetailModal. (and change imports too)

jmmxp:
For LoggedInUserContext, you can comment `Allows App to globally access the logged in user after login/registration. App then sends UserLogin through props to child components.`

front-end/src/components/PostListItem.tsx
```diff
@@ -13,6 +13,10 @@ interface Props {
   removeFromFeed: Function;
 }
 
+/**
+ * Post list item component to show a preview for an individual post on a stream
```
VinnyLuu, line 18: I think the PostContentEl is the preview. The comment could be like: "PostListItem component is an individual post list item on a stream" or something like this.

VinnyLuu (approved):
LGTM made a comment about one of the comments hehe. You can push after addressing this and any other reviews

jmmxp (approved):
Sorry, I'm trying to get a more succinct comment for what LoggedInUserContext is for. Last change, I swear: 

`Allows App to globally access the logged in user after login/registration. App then sends UserLogin through props to child components.`

PR #112, opened 2021-03-05
Fix error on ordering posts

VinnyLuu:
With @Jmmxp , we changed ordering to be on the model, adding a meta class to Post to sort instead of sorting on the queryset.


jmmxp:
The bug is that my line of code should have been `queryset = queryset.order_by`, but we were ignoring the return value. Too used to Python's sorting methods that don't return anything.

jmmxp (approved):
LGTM

PR #120, opened 2021-03-08
View Other Author's Pages

### LINKED ISSUE #96: View Other Author's Profiles
Getting an access denied when trying to go to another author's profile.

Not necessarily part of any issue, but is more-so related to getting author's friends' posts

raysarinas:
Closes #96 

- Fix not being able to view other author's pages. 
- When you view another author's page, you can't see their stream (unauthorized request when you try, so right now you can only see your own posts since you're authenticated to do so). 
- Introduces weird bug where if you're on someone else's author page, and then try to navigate to your own profile (clicking the "Profile" button on the navbar), you are redirected to the route with your ID but contents are not updated. 

What your profile looks like:
Screen Shot 2021-03-07 at 5 01 58 PM

What other profiles look like:
Screen Shot 2021-03-07 at 4 49 21 PM

### DIFF
--- a/front-end/src/pages/AuthorPage.tsx
+++ b/front-end/src/pages/AuthorPage.tsx
@@ -40,7 +40,7 @@ export default function AuthorPage(props: any) {
     })
 
     // Only get stream if you're viewing your own profile
-    if (props.loggedInUser && props.location.pathname.includes(props.loggedInUser.authorId)) {
+    if (props.loggedInUser) {
       axios.get(authorUrl + "/posts/",
         {
           auth: { // authenticate the GET request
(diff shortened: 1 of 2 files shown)


front-end/src/pages/AuthorPage.tsx
```diff
@@ -90,7 +91,7 @@ export default function AuthorPage(props: Props) {
             </CardBody>
           </Card>
         </Col>
-        {<PostList postEntries={postEntries} setPostEntries={setPostEntries} loggedInUser={props.loggedInUser} />}
+        {author && displayPosts()}
```
ChrisChrisLoLo, line 94: I like the use of this conditional

ChrisChrisLoLo (commented):
Code LGTM!

I think we might want to display posts of other authors though (so long as their stuff isn't private or unlisted). That way we can like and comment on their posts

front-end/src/pages/AuthorPage.tsx
raysarinas: ty, fixed/added viewing other authors public stream. 

the only issue is that if we're not logged in and we try to view an author's profile, there is an authentication issue that I don't know where its coming from (probably on the backend) so it wont GET the author's posts. for now, I have it so that posts only get displayed if you're logged in and viewing an author's profile

ChrisChrisLoLo (approved):
LGTM

PR #122, opened 2021-03-08
Persist user login

ChrisChrisLoLo:
User login will now persist in LocalStorage.

You can now refresh the page without having to re-login


raysarinas (approved):
LGTM. Not really sure what the code does, so maybe someone else can also take a look at it, but I don't see any problems with it! 💯

PR #125, opened 2021-03-10
Add Netlify configuration

ChrisChrisLoLo:
Allows us to deploy the front-end with a fair amount of ease.

We can use a service (Netlify) to automatically publish our front-end with every commit. The current drawback is we need different `.env` files depending if we're production or development. For now we can do this manually.

We can talk about adding more automation/scripts later on.

Changes:

- added config that will allow netlify to automatically build from our git
- added options to Django to open CORS up to everything

### DIFF
--- a/back-end/api/api/settings.py
+++ b/back-end/api/api/settings.py
@@ -146,7 +146,4 @@ REST_FRAMEWORK = {
     'PAGE_SIZE': 10,
 }
 
-CORS_ALLOWED_ORIGINS = [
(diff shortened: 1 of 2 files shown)


ChrisChrisLoLo:
An example of a deployment (should work if you have your local Django running) https://604952270547080007261cd0--serene-shaw-3f9747.netlify.app/

ChrisChrisLoLo:
I'll merge this in in a few days if no reviews come in

PR #128, opened 2021-03-13
Post detail page

ChrisChrisLoLo:
Create post detail page

Adjust margins on home page

Detail page:
Screenshot_20210312_232937

New margins:
Screenshot_20210312_232951

### DIFF
--- a/front-end/src/pages/PostDetailPage.tsx
+++ b/front-end/src/pages/PostDetailPage.tsx
@@ -1,0 +1,62 @@
+import axios, { AxiosResponse } from 'axios';
+import React, { useEffect, useState } from 'react';
+import { RouteComponentProps } from 'react-router';
+import { useParams } from 'react-router-dom';
+import { Row, Col } from 'reactstrap';
+import PostDetailModal from '../components/PostDetailModal';
+import { Post } from '../types/Post';
+import { UserLogin } from '../types/UserLogin';
+
+
+interface Props extends RouteComponentProps<MatchParams>{
+  loggedInUser: UserLogin|undefined;
+}
+
+interface MatchParams {
+  postId: string
+}
+
+/**
+ * Renders a post detail
+ * @param props 
+ */
+export default function PostDetailPage(props: Props) {
+  const [postEntry, setPostEntry] = useState<Post | undefined>(undefined);
+
+  const {postId} = props.match.params;
+  // get post
+  useEffect(() => {
+    let getPromise: Promise<AxiosResponse<any>>;
+    if (props.loggedInUser) {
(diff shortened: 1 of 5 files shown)


front-end/src/pages/PostDetailPage.tsx
```diff
@@ -0,0 +1,62 @@
...
+      getPromise = axios.get(`${process.env.REACT_APP_API_URL}/api/posts/${postId}`,
+        {
+          auth: {
+            username: props.loggedInUser.username,
+            password: props.loggedInUser.password
+          }
+        });
+    }
+    else {
+      getPromise = axios.get(`${process.env.REACT_APP_API_URL}/api/posts/${props.location.pathname}`,);
```
jmmxp, line 40: Just to make sure I understand 100%, what's the difference between making the request w/ Basic Auth and w/o?

jmmxp (approved):
LGTM, just had a quick question

raysarinas:
> Detail page:
> Screenshot_20210312_232937

Do you just get redirected to the stream/homepage after you close the post details page? I guess the background behind the modal in this screenshot looks funny to me? Could we embed the modal into a page so we can still use the navbar while viewing post details? Not sure if it's possible, or if its something we could hold off on until the very end as something extra

front-end/src/pages/PostDetailPage.tsx
ChrisChrisLoLo: Good question :+1:

This logic is to allow for logged-in users and non-logged-in users to attempt to access a post.
I'm assuming that some friends-only posts need user credentials in order to determine if the server should return the post or not

I'm assuming that the server will refuse to send a friends-only post if 'bad' credentials or no credentials get sent to it

There might need to be additional work to get this edge case completed though

ChrisChrisLoLo:
> > Detail page:
> > Screenshot_20210312_232937
> 
> Do you just get redirected to the stream/homepage after you close the post details page? I guess the background behind the modal in this screenshot looks funny to me? Could we embed the modal into a page so we can still use the navbar while viewing post details? Not sure if it's possible, or if its something we could hold off on until the very end as something extra

Valid concern. You do get redirected when closing the modal. It'd be fairly tricky to embed the modal into a page since IIRC React Router only supports rendering individual pages, not individual components that can be added onto pages. 

There might be other tricks we can do that are fairly simple (give the page a nice texture, make the modal shadow darker, anything that can make the plain background less apparent), so it would be nice to have a ticket for this so we can polish it up at the end. Making the component seamless with the homepage could be fairly non-trivial though, or may involve some hacks/code-duplication

front-end/src/pages/PostDetailPage.tsx
jmmxp: Got it, makes sense. Thanks for clarifying!

Issue #127, opened 2021-03-11
Add Follow/Friend Requests UI Page


raysarinas:
### Related User Stories
**36. As an author, I want to know if I have friend requests**

### Description / Notes
- New page (link in navbar?) to a page that lists your friend/follow requests
- Have a GET request sent to get your friend/follow requests from your inbox

PR #129, opened 2021-03-14
Follow/Friend Request UI

raysarinas:
Related to #127 

A modal to display who has followed you, using the `/api/author/{AUTHOR_ID}/inbox` endpoint. Link to open the modal is currently in the navbar, but this can be changed to be on the main feed page and/or your own author profile page. 

Friend/follow requests list people who have followed you but you have not followed back.

`Befriend` button should follow back who has followed you.

**TODO**: After you befriend/follow back, the list should update. We can have this either filter out items from our inbox based on who we follow (i.e., if we have a following list, just check to see if the person is in both our following and followers list) or create an endpoint to remove the request from our inbox. 

Screen Shot 2021-03-14 at 3 32 17 PM


front-end/src/components/FriendRequestItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default function FriendRequestItem(props: Props) {
+  const friendRequest = props.friendRequest;
+
+  function followBack(requester: Author) {
+    // TODO: follow back
+    console.log("followed back: " + requester.displayName);
+  }
+
+  return (
+    // <>
```
VinnyLuu: remove comment

front-end/src/components/FriendRequestListModal.tsx
```diff
@@ -0,0 +1,42 @@
+import axios from 'axios';
+import React, { useEffect, useState } from 'react';
+import { Button, Card, CardBody, CardTitle, Modal, ModalBody, ModalHeader } from 'reactstrap';
+import { Follow } from '../types/Follow';
+import FriendRequestItem from './FriendRequestItem';
+
+const FriendRequestListModal = (props: any) => {
```
VinnyLuu, line 7: a props interface would be nice here just so that its clear what props are being passed in

VinnyLuu (approved):
LGTM just left a couple comments

front-end/src/components/FriendRequestListModal.tsx
raysarinas: I tried that at first, but it gets really messy with having to deal with the `UserLogin` object being defined or undefined. I imagine there might be a similar issue in `CreatePost.tsx` which does the same thing since it also just passes in `props: any`

front-end/src/components/FriendRequestItem.tsx
raysarinas: done

PR #130, opened 2021-03-15
Use partial_update for Post editing in DRF

jmmxp:
We had a bug where posts could not be edited unless every field was sent. (e.g., editing a post that didn't have a description would result in an error). Fixed this by changing DRF to use partial_update instead of just update.

Added tests for this and also for author retroactively.


VinnyLuu (approved):
LGTM

PR #131, opened 2021-03-15
Update Like model, update Inbox endpoint to save Like objects 

jmmxp:
I changed the Like model's author field to be a JSONField, since we plan to store author objects exactly as given in the Like.  Changes to corresponding `Like` tests made to tests to ensure they still pass.

After that, I added code to the Inbox endpoint, which checks the type of the object given, and if it is a like, creates a Like object into our database. @ChrisChrisLoLo should now be able to send Likes to inboxes, and then retrieve them later for a post/comment by GET'ing it.


VinnyLuu (approved):
LGTM

PR #133, opened 2021-03-20
Find likes by checking for substring (post/comment ID) in object field 

jmmxp:
Previously, this wouldn't have worked properly because if the "object" field of the Like actually stores an entire URL, and the endpoint `api/author/<str:author>/posts/<str:post>/likes/` would break unless <str:post> was an entire URL, which would still also break unless we encode the URL before sending.

To fix this when people ask for likes for a post/comment, the <str:post> should just be a UUID. Thus, we find all likes where its object field contains this UUID substring.


VinnyLuu (approved):
LGTM

PR #135, opened 2021-03-20
Send Friend Request to Inbox

raysarinas:
Used Vinny's follow button logic and used it in the friend request modal. No changes to UI other than that the button just says Follow/Unfollow now.

Screen Shot 2021-03-20 at 1 35 52 AM
Screen Shot 2021-03-20 at 1 36 00 AM

Cleaned up search page so results look like this now: 
Screen Shot 2021-03-20 at 1 29 35 AM

Also changed the feed names so it's clearer/less confusing:
image


front-end/src/components/FriendRequestButton.tsx
```diff
@@ -11,21 +11,36 @@ interface Props {
...
+          summary: "", // do we need to make custom messages when we follow? prob not
+          actor: res.data,
+          object: props.currentAuthor
+        };
+
+        axios.post(inboxUrl, follow).then(res => {
+          if (res.status === 201) {
+            console.log("friend request sent");
+          }
+        });
```
VinnyLuu: This should be moved to the above if statement so that it sends the follow message to the inbox only when we get a 201

VinnyLuu (changes requested):
LGTM after the one comment. Thanks for putting in the follow button for the follow back.

front-end/src/components/FriendRequestButton.tsx
raysarinas: done

VinnyLuu (approved):

PR #136, opened 2021-03-20
Add 'api/' and trailing slash to author url

jmmxp:
This makes it so that this author URL can be directly GET'd to retrieve the author information.


VinnyLuu (approved):
LGTM

PR #137, opened 2021-03-20
Remove unused context and summary fields on Like model

jmmxp:
Discussed with other groups that we don't need these fields.


VinnyLuu (approved):
LGTM

PR #138, opened 2021-03-20
Match authors/ endpoint with other teams

jmmxp:
We had some pagination attributes etc. from the router default endpoint. I've added our own ViewSet to return all authors now and added proper tests as well. 

Result looks like this now:
```javascript
[
    {
        "id": "f9e8eb9e-f30c-44d6-bc1c-a1450543b389",
        "type": "author",
        "displayName": "abc",
        "url": "http://localhost:8000/api/author/f9e8eb9e-f30c-44d6-bc1c-a1450543b389/",
        "github": "abc",
        "host": "http://localhost:8000/"
    },
    {
        "id": "93721ae8-1d9a-4f07-b1d3-d5e640250616",
        "type": "author",
        "displayName": "abc2",
        "url": "http://localhost:8000/api/author/93721ae8-1d9a-4f07-b1d3-d5e640250616/",
        "github": "abc2",
        "host": "http://localhost:8000/"
    }
]
```

### DIFF
--- a/back-end/api/quickstart/tests/endpoints/test_authors.py
+++ b/back-end/api/quickstart/tests/endpoints/test_authors.py
@@ -1,0 +1,27 @@
+import json
+from rest_framework import status
+from django.test import TestCase, Client
+from quickstart.models import Author
+from quickstart.serializers import AuthorSerializer
+from quickstart.tests.helper_test import get_test_author_fields
+
+client = Client()
+
+class GetAuthorById(TestCase):
+  """Tests for getting all Authors at endpoint /api/authors/."""
+  def test_get_authors(self):
+    for i in range(3):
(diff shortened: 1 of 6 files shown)


back-end/api/quickstart/tests/endpoints/test_authors.py
```diff
@@ -0,0 +1,27 @@
+import json
```
VinnyLuu: rm unused import

VinnyLuu (approved):
LGTM

back-end/api/quickstart/tests/endpoints/test_authors.py
jmmxp: Done

PR #140, opened 2021-03-21
Send Follows to Inboxes upon Follow creation

jmmxp:
I changed the Inbox model to use a FOREIGN_KEY to Author objects, and updated tests to reflect this. I then added code to the Create Follow endpoint that ensures that the follow object is sent to the inbox of whoever the follow was sent to.

a Follow (Friend Request) object stored in the inbox looks as follows: (actor is following object)
image


back-end/api/quickstart/tests/helper_test.py
```diff
@@ -1,4 +1,5 @@
 import datetime
+import json
```
VinnyLuu: Nit remove

VinnyLuu (approved):
LGTM

back-end/api/quickstart/tests/helper_test.py
jmmxp: Done

PR #141, opened 2021-03-21
Fixed failing test for follows

VinnyLuu:
The GETing a follower test was failing due to the serializer JSON changes I had made previously. This test was failing because the order of properties on the response was not matching that of the serializer. I can take more time to look into it if we want but right now I don't see an issue with the properties not being in the same order and have changed the test to just make sure that all fields are correct. 

Added additional error checking on the view if the author searched for follows does not exist. 

Changed retrieve to be using get instead of filter when searching since there is only one author that should match.

### DIFF
--- a/back-end/api/quickstart/views.py
+++ b/back-end/api/quickstart/views.py
@@ -206,13 +206,11 @@ class FollowersViewSet(MultipleFieldLookupMixin, viewsets.ModelViewSet):
     def destroy(self, request, receiver, sender):
         try:
             author = Author.objects.get(id=receiver)
-            queryset = Follow.objects.filter(receiver=author, sender__id=sender)
-            if len(queryset) == 0:
-                return Response(status=status.HTTP_404_NOT_FOUND)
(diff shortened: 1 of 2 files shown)


back-end/api/quickstart/views.py
```diff
@@ -199,14 +204,17 @@ def create(self, request, receiver, sender):
...
     def destroy(self, request, receiver, sender):
-        author = Author.objects.get(id=receiver)
-        queryset = Follow.objects.filter(receiver=author, sender__id=sender)
-        if len(queryset) == 0:
-            return Response(status=status.HTTP_404_NOT_FOUND)
+        try:
+            author = Author.objects.get(id=receiver)
+            queryset = Follow.objects.filter(receiver=author, sender__id=sender)
+            if len(queryset) == 0:
+                return Response(status=status.HTTP_404_NOT_FOUND)
```
jmmxp: use .get and except Follow.DoesNotExist here as well?

jmmxp (approved):
LGTM

PR #147, opened 2021-03-23
Add Node model and endpoint to get all Nodes

jmmxp:
A node represents an external server we can make requests to.

I will make a follow-up PR to get all these nodes in the front-end upon user login/registration.

The nodes can then be used whenever we make requests: instead of `axios.get`, we will do `axioswrapper.get` which checks the URL the request is being made to and uses the proper credentials to make the request.


VinnyLuu (approved):
LGTM

PR #152, opened 2021-03-23
Followers List UI

### LINKED ISSUE #151: List of Followers UI
### Related User Stories
**34. As an author, my server will know about my friends**

raysarinas:
Closes #151 

Without a doubt, an issue on the frontend / react router. Requests to Django are being made properly, but navigating to `/author/author_id/followers/` breaks. 

For now, we have the url/route as `/followers/ABC/`, where `author_id = ABC` is the author and we are seeing their list of followers. I will open up a separate issue for fixing this so that we can navigate to `/author/ABC/followers/` instead

Followers Button on profile page:
Screen Shot 2021-03-22 at 6 36 43 PM

Followers list page:
Screen Shot 2021-03-22 at 6 11 08 PM


jmmxp:
Just want to double check - the "Unfollow" button shouldn't show unless it's someone we're "Following" right?

raysarinas:
> Just want to double check - the "Unfollow" button shouldn't show unless it's someone we're "Following" right?

Yeah, it will basically just toggle between saying "Follow" and "Unfollow", like if I unfollow then the button looks like this:

Screen Shot 2021-03-22 at 6 46 07 PM

front-end/src/pages/FollowersPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Row, Col, Card, CardBody, CardTitle, Container } from 'reactstrap';
+import AuthorList from "../components/AuthorList"
+import AuthorListItem from '../components/AuthorListItem';
+import { Author } from '../types/Author';
+
+/**
+ * Render list of search results when searching for an author by display name
+ * @param props 
+ */
+export default function FollowersPage(props: any) {
```
ChrisChrisLoLo: Try to type props if you can

https://github.com/CMPUT404W21T02/group-project/blob/main/front-end/src/pages/PostDetailPage.tsx might be a good reference

ChrisChrisLoLo (commented):
I think I have a guess as to why the author page is acting up:
https://stackoverflow.com/questions/54986386/react-router-4-and-exact-path-with-dynamic-param

If you want to try fixing the routing so that you get /authors/.../followers, it might work if you put the `<Route>` component above the existing one, so react router doesn't render the `AuthorPage` component

front-end/src/pages/AuthorPage.tsx
```diff
@@ -41,16 +43,18 @@ export default function AuthorPage(props: any) {
...
     if (props.loggedInUser) {
-      // get whether user is follower of author
-      axios.get(authorUrl + "/followers/" + props.loggedInUser.authorId).then(res => {
-        setIsFollower(true);
-      }).catch(err => {
-        // 404 is not a follower
-        setIsFollower(false);
-      })
+      // get whether user is follower of author IF not looking at our own profile
+      if (!props.location.pathname.includes(props.loggedInUser.authorId)) {
```
ChrisChrisLoLo, line 50: In the future we might want to avoid using `props.location.pathname`. React router has a way for us to get the exact parameters we want as props. 

Here is an example of how to do it (note the `MatchParams` interface): https://github.com/CMPUT404W21T02/group-project/blob/main/front-end/src/pages/PostDetailPage.tsx

https://stackoverflow.com/questions/48138111/what-typescript-type-should-i-use-to-reference-the-match-object-in-my-props

ChrisChrisLoLo (approved):
LGTM!

A few comments, but I think this is good to merge in

front-end/src/pages/FollowersPage.tsx
raysarinas: done

front-end/src/pages/AuthorPage.tsx
raysarinas: added a TODO comment. I think we can fix this when we get to work on the axios wrapper, since it would involve us having to change paths/URLs at that point

PR #153, opened 2021-03-23
Retrieve nodes on front-end upon login and registration

jmmxp:
Upon author login/registration, they will receive all nodes that are currently in our back-end.

The nodes still need to be used when making requests, but this can be done by me/Chris in a future PR. Will require making an axios wrapper. Also we have the decision of using this wrapper everywhere or only for the requests that we might make to external servers.

### DIFF
--- a/front-end/src/App.tsx
+++ b/front-end/src/App.tsx
@@ -37,8 +37,8 @@ function App() {
     }
   },[loggedInUser])
 
-  // TODO wrap the below in NodesContext as well, and then send it into pages and use it 
-  // in the Axios wrapper. OR make Nodes global.
+  // TODO wrap the below in NodesContext as well, and then use the Nodes in other components.
+  // Talk to Chris about how to use React Context.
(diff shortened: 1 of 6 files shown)


front-end/src/App.tsx
```diff
@@ -36,16 +37,18 @@ function App() {
     }
   },[loggedInUser])
 
+  // TODO wrap the below in NodesContext as well, and then send it into pages and use it 
+  // in the Axios wrapper. OR make Nodes global.
```
VinnyLuu: do you need this TODO since NodesContexts makes it global?

VinnyLuu (approved):
LGTM

front-end/src/App.tsx
jmmxp: good question, I'm not 100% familiar with how React Context works. I've updated the comment to not say to pass it around but to get in touch with Chris about how to do this.

PR #155, opened 2021-03-23
Deploy backend

Monica-Uwalaka:


raysarinas (commented):
LGTM, but we already have a `requirements.txt` in the `back-end` directory. Now there are three `requirements.txt` files. Do we need all three? Or could we just keep one of them?

Monica-Uwalaka:
> LGTM, but we already have a `requirements.txt` in the `back-end` directory. Now there are three `requirements.txt` files. Do we need all three? Or could we just keep one of them?

We need one in the root directory, We can keep the current one if there are no conflicts with the one I just pushed. I don't know if we also need one in the back-end folder, I wasn't the one that added the one in the back-end. But I am guessing we need it when we run manage.py locally. So I'll leave it and we'll have two.

back-end/api/api/settings.py
```diff
@@ -26,7 +26,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ["https://cyver.herokuapp.com", 'http://127.0.0.1']
```
ChrisChrisLoLo: Can we make this line like this?
```suggestion
ALLOWED_HOSTS = ["*"]
```
NOTE: I haven't tested this.

https://docs.djangoproject.com/en/3.1/ref/settings/#allowed-hosts

Currently I'm not sure if other's deployment will work if we only allow for "https://cyver.herokuapp.com" and localhost, since we'll have different domains.

Making it "*" _should_ allow anyone to host it regardless of what our Heroku hostname is. It's not secure, but for the sake of the project this should be fine.

Could you see if "*" works?

back-end/api/requirements.txt
```diff
@@ -0,0 +1,29 @@
+asgiref==3.3.1
+certifi==2020.12.5
+cffi==1.14.4
```
ChrisChrisLoLo, line 3: I had problems with the requirements.txt too when I dabbled getting it to work with Heroku, so as long as the app deploys automatically I'm personally good with it. We just need to be careful as a team if we install new packages

ChrisChrisLoLo (changes requested):
Good stuff! Thanks for the fast rollout of this.

Could you try my first comment and see if it works? If we use "*" hopefully the server should deploy regardless of host name

back-end/api/api/settings.py
Monica-Uwalaka: Yes I'll do it now

back-end/api/requirements.txt
Monica-Uwalaka: Sounds good!

back-end/api/api/settings.py
Monica-Uwalaka: It works

ChrisChrisLoLo (approved):
LGTM

Issue #148, opened 2021-03-23
Frontend Axios Wrapper


raysarinas:
### Related User Stories
**28. As a server admin, I don't want to do heavy setup to get the posts of my author's friends
**

Rough example from Justin:
```
wrapper.get(url) // use these instead of axios
wrapper.post(url) 

function get(url) {
  for each key in nodes
    if url contains(key)
      make the request using credentials
}
```

jmmxp:
Done, wrapper was created and used in #167, and request URLs were updated in #174

jmmxp:
Made updates to the wrapper in #182.

PR #158, opened 2021-03-24
Following List UI

raysarinas:
Following list is done. Followers/following lists are on one page as tabs, but the route/URL will redirect to `/author/author_id/following` and `/author/author_id/followers` as you change the tabs. Only thing left to do is to get the `following` API endpoint going and uncomment an axios request that is in the code. I guess it is also possible to get the list of who you follow without that custom endpoint, but probably better to have that endpoint.

BTW: also changed the colours on the Followers/Following tabs to grey. Screenshots just show it as blue

On profile page:
Screen Shot 2021-03-23 at 5 55 48 PM

Following:
Screen Shot 2021-03-23 at 5 56 03 PM

Followers:
Screen Shot 2021-03-23 at 5 56 25 PM


front-end/src/pages/AuthorPage.tsx
```diff
@@ -100,9 +100,16 @@ export default function AuthorPage(props: any) {
...
-  const displayFollowersListButton = () => {
+  const displayFollowListButtons = () => {
     if (props.loggedInUser && author?.id === props.loggedInUser.authorId) {
-      return <Button><CardLink className="text-white" href={"/author/" + props.loggedInUser.authorId + "/followers"} >Followers</CardLink></Button>
+      return (<>
+        <CardText>
+          <Button><CardLink className="text-white" href={"/author/" + props.loggedInUser.authorId + "/followers"}>Followers</CardLink></Button>
+        </CardText>
+        <CardText>
+        <Button><CardLink className="text-white" href={"/author/" + props.loggedInUser.authorId + "/following"}>Following</CardLink></Button>
```
VinnyLuu: nit: tab

front-end/src/components/FriendRequestButton.tsx
```diff
@@ -26,12 +26,22 @@ export default function FollowRequestButton(props: Props) {
             props.setIsFollower(true);
           }
         });
+        axios.put(followingUrl, props.currentAuthor).then(res => {
+          if (res.status === 201) {
+            console.log(props.loggedInUser?.username + " is now following " + props.currentAuthor?.displayName);
+          }
+        })
```
VinnyLuu: nit: semicolon

front-end/src/pages/FollowListPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </Nav>
+      </Row>
+      <Row className="justify-content-md-center">
+        <Col sm="12" md={{ size: 6 }}>
+          <TabContent activeTab={activeTab}>
+            <TabPane tabId="followers">
+              <Row>
+                <Col sm="12">
+                  {followers?.length !== 0 ?
+                    followers?.map((follower: Author) => <AuthorListItem author={follower} loggedInUser={props.loggedInUser}></AuthorListItem>) :
```
VinnyLuu: nit: you could refactor the mapping for both followers and following to a method

front-end/src/App.tsx
```diff
@@ -50,10 +50,11 @@ function App() {
             <Switch>
               <Route exact path="/" render={(props) => <HomePage {...props} loggedInUser={loggedInUser} />} />
               <Route path="/auth" render={(props) => <AuthPage {...props} setLoggedInUser={setLoggedInUser} setNodes={setNodes}/>} />
-              <Route path="/author/:authorId/followers/" render={(props) => <FollowersPage {...props} loggedInUser={loggedInUser} />}/>
+              <Route path="/author/:authorId/followers/" render={(props) => <FollowListPage {...props} loggedInUser={loggedInUser} activeTab={'followers'} />}/>
+              <Route path="/author/:authorId/following/" render={(props) => <FollowListPage {...props} loggedInUser={loggedInUser} activeTab={'following'}/>}/>
               <Route path="/author/:authorId" render={(props) => <AuthorPage {...props} loggedInUser={loggedInUser}/>}/>
               {/* TODO: hide settings page if not logged in */}
-              <Route path="/settings" render={(props) => <SettingsPage loggedInUser={loggedInUser} />} />
+              <Route path="/settings" render={() => <SettingsPage loggedInUser={loggedInUser} />} />
```
VinnyLuu, line 86: did it ever need those props?

front-end/src/components/FriendRequestButton.tsx
```diff
@@ -26,12 +26,22 @@ export default function FollowRequestButton(props: Props) {
...
         axios.delete(authorUrl).then(res => {
           if (res.status === 204) {
             props.setIsFollower(false);
           }
         });
+        axios.delete(followingUrl).then(res => {
+          if (res.status === 204) {
+            console.log("UNFOLLOWED");
+          }
+        })
```
VinnyLuu: nit: semicolon

VinnyLuu (approved):
LGTM

front-end/src/App.tsx
raysarinas: no, that's why I took it out

front-end/src/components/FriendRequestButton.tsx
```diff
@@ -13,8 +13,8 @@ interface Props {
 export default function FollowRequestButton(props: Props) {
   // TODO: change this to use currentAuthor.url
   const authorUrl = process.env.REACT_APP_API_URL + "/api/author/" + props.currentAuthor?.id + "/followers/" + props.loggedInUser?.authorId + "/";
-  const inboxUrl = process.env.REACT_APP_API_URL + "/api/author/" + props.currentAuthor?.id + "/inbox/";
   const loggedInUserUrl = process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser?.authorId;
+  const followingUrl = loggedInUserUrl + "/following/" + props.currentAuthor?.id + "/";
```
ChrisChrisLoLo: This is unrelated, but it might be good to start using format strings in the future:
https://stackoverflow.com/questions/610406/javascript-equivalent-to-printf-string-format

front-end/src/components/FriendRequestButton.tsx
```diff
@@ -26,12 +26,22 @@ export default function FollowRequestButton(props: Props) {
             props.setIsFollower(true);
           }
         });
+        axios.put(followingUrl, props.currentAuthor).then(res => {
+          if (res.status === 201) {
+            console.log(props.loggedInUser?.username + " is now following " + props.currentAuthor?.displayName);
```
ChrisChrisLoLo, line 30: Should we print some sort of warning/error if the loggedInUser is undefined?

front-end/src/pages/FollowListPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+              <h2>Following</h2>
+            </NavLink>
+          </NavItem>
+        </Nav>
+      </Row>
+      <Row className="justify-content-md-center">
+        <Col sm="12" md={{ size: 6 }}>
+          <TabContent activeTab={activeTab}>
+            <TabPane tabId="followers">
+              <Row>
```
ChrisChrisLoLo, line 79: The positioning might be a bit off if we don't wrap `<Row>` els with `<Containter>`

front-end/src/pages/FollowListPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    followers?.map((follower: Author) => <AuthorListItem author={follower} loggedInUser={props.loggedInUser}></AuthorListItem>) :
+                    <Card body className="text-center"><CardBody><CardTitle tag="h5" >You have no followers :(</CardTitle></CardBody></Card>
+                  }
+                </Col>
+              </Row>
+            </TabPane>
+            <TabPane tabId="following">
+              <Row>
+                <Col sm="12">
+                  {following?.length !== 0 ?
```
ChrisChrisLoLo: We might want to use the optional chaining operator less frequently going forward. There could potentially be bugs that pop-up if `following` becomes undefined for a given reason. 

We may want to go for something like this
```ts
if (following) {
    // do thing
    console.log(following.length)
} else {
    // warn here
    console.error('Following is undefined when it shouldn\'t be!')
}
```
As it'll notify us if our assumptions about `following` existing ever breaks

ChrisChrisLoLo (approved):
Looks good! Thanks for doing this!

The comments I left are suggestions for potential code going forward

front-end/src/components/FriendRequestButton.tsx
raysarinas: Sounds good, I think anything related to URLs and axios requests we should address when we get to making the wrapper for it (Issue #148)

front-end/src/pages/FollowListPage.tsx
raysarinas: The whole page is already in a container

ChrisChrisLoLo: Correct, though currently, iirc it's
```jsx
<Container>
    <Row>
        <Col>
           <Row>
                ...
           </Row>
        </Col>
    </Row>
</Container>
```
we should have it like this, or else the CSS can act up, since we usually don't put `<Row>` elements directly in `<Col>` elements
```jsx
<Container>
    <Row>
        <Col>
           <Container>
               <Row>
                ...
               </Row>
           </Container>
        </Col>
    </Row>
</Container>
```

raysarinas: fixed

PR #159, opened 2021-03-24
Implement following models and endpoint

VinnyLuu:
Implemented the following model and endpoints to add people an author on our server is following. For example, Author A on our server sent a follow request to Author B (can be either on our server or another server, which is why it is JSON and also why it is separate from the Follow model), this model will have Author A as a foreign key and Author B as JSON. 

Also changed the name of get_sender_fields helper method.

Passed all tests.


jmmxp (approved):
LGTM

PR #160, opened 2021-03-24
Keep track of Node for our own server upon login/registration

jmmxp:
I've serialized the requests on Login/Registration (and also removed many redundant `.catch` expressions on Promises) so that it goes
POST login/register ->
GET nodes ->
GET authuser

When GET authuser succeeds, I've added code into its useEffect to add a Node for our own server into the Nodes list. Then the axios wrapper we create can properly make requests to our own server as well.

EDIT: I've also added code to save these nodes to local storage so they aren't lost on refresh. Uses the same mechanism as UserLogin does.

Here's what Node objects look like.
image

### DIFF
--- a/front-end/src/App.tsx
+++ b/front-end/src/App.tsx
@@ -8,6 +8,7 @@ import NotFoundPage from './pages/NotFoundPage';
 import AuthPage from './pages/AuthPage';
 import { LoggedInUserContext } from './contexts/LoggedInUserContext';
 import { UserLogin } from './types/UserLogin';
+import { Node } from './types/Node';
 import AuthorPage from './pages/AuthorPage';
 import SettingsPage from './pages/SettingsPage';
 import CreatePostComponent from './components/CreatePost';
@@ -36,6 +37,15 @@ function App() {
     } else {
       localStorage.setItem(LOCAL_STORAGE_USER,JSON.stringify(loggedInUser));
     }
+
+    if (loggedInUser?.username !== undefined && loggedInUser?.password !== undefined) {
+      // Keep track of a Node representing our own server
+      setNodes(n => [...n, {
+        "host": process.env.REACT_APP_API_URL, // TODO may need to strip http:// from this host.
+        "username": loggedInUser?.username, 
+        "password": loggedInUser?.password
+      } as Node])
+    }
   },[loggedInUser])
 
   // TODO wrap the below in NodesContext as well, and then use the Nodes in other components.
@@ -45,7 +55,7 @@ function App() {
       <LoggedInUserContext.Provider value={loggedInUser}>
       <BrowserRouter>
         <div>
-          <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} />
+          <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} setNodes={setNodes} />
           <Container fluid={true}>
             <Switch>
               <Route exact path="/" render={(props) => <HomePage {...props} loggedInUser={loggedInUser} />} />
--- a/front-end/src/App.tsx
+++ b/front-end/src/App.tsx
@@ -17,6 +17,7 @@ import PostDetailPage from './pages/PostDetailPage';
 import FollowersPage from './pages/FollowersPage';
 
 const LOCAL_STORAGE_USER = 'loggedInUser';
+const LOCAL_STORAGE_NODES = 'nodes';
 
 /*
 * Snippet based on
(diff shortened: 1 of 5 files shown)


front-end/src/App.tsx
```diff
@@ -24,19 +26,46 @@ const LOCAL_STORAGE_USER = 'loggedInUser';
...
-  const initialJSON = localStorage.getItem(LOCAL_STORAGE_USER);
-  const initialState:UserLogin|undefined = initialJSON ? JSON.parse(initialJSON) : undefined;
+  const initialUserJSON = localStorage.getItem(LOCAL_STORAGE_USER);
+  const initialUserState: UserLogin|undefined = initialUserJSON ? JSON.parse(initialUserJSON) : undefined;
+  const [loggedInUser,setLoggedInUser] = useState<UserLogin | undefined>(initialUserState);
 
-  const [loggedInUser,setLoggedInUser] = useState<UserLogin | undefined>(initialState);
-  const [nodes, setNodes] = useState<Node[]>([]);
+  const initialNodesJSON = localStorage.getItem(LOCAL_STORAGE_NODES);
+  const initialNodesState: Node[] = initialNodesJSON ? JSON.parse(initialNodesJSON) : [];
```
ChrisChrisLoLo, line 34: Awesome that initialNodesState gets initialized as `[]` rather than `undefined`

front-end/src/App.tsx
```diff
@@ -45,7 +74,7 @@ function App() {
       <LoggedInUserContext.Provider value={loggedInUser}>
       <BrowserRouter>
         <div>
-          <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} />
+          <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} setNodes={setNodes} />
```
ChrisChrisLoLo, line 77: This is good :+1:

For the sake of communicating this now, we should only let devs pass in `setNodes` as a prop, but never `nodes` itself. This way we can hopefully drill down props a bit less

front-end/src/App.tsx
```diff
@@ -24,19 +26,46 @@ const LOCAL_STORAGE_USER = 'loggedInUser';
...
+      // Already have a Node for our own server.
+      if (nodes.filter(n => n.username === loggedInUser.username).length !== 0) {
+        return;
+      }
+
+      // Strip http:// or https:// from our own API URL
+      const host = process.env.REACT_APP_API_URL?.replace("http://", "").replace("https://", "");
+
+      // Keep track of a Node representing our own server
+      setNodes([...nodes, {
```
ChrisChrisLoLo, line 52: Out of curiosity, is there a case for replacing server credentials with our own?

We can probably this as is either way, but I'm not sure if I'm forgetting an edge case where this part is required

front-end/src/components/LoginForm.jsx
```diff
@@ -36,26 +36,16 @@ export default class LoginForm extends React.Component {
...
-      });
-
-      axios.get(process.env.REACT_APP_API_URL + "/api/nodes/").then(res => {
-        this.props.setNodes(res.data);
-      }).catch(err => {
-        this.setState({loginErr: true});
-      });
-
+    }).then(_ => {
+      return axios.get(process.env.REACT_APP_API_URL + "/api/nodes/");
```
ChrisChrisLoLo, line 40: Off-topic: Is there a convention guide for this style of promise chaining? Personally, I find this a bit uncommon and harder to read as opposed to nested callbacks, since the `return` statements aren't breaking out of the function. I might be in the minority though

We should leave this as is, since either way I'm good with it, but if I'm writing code that's not idiomatic I'd like to improve in that area

ChrisChrisLoLo (approved):
LGTM :shipit: 

Comments mostly are questions

Out of curiosity, is there a reason why an unauthenticated user can't have `nodes`? Could we instead just always get the nodes when we open up the website? Either way is fine in my head, but I'd just like to confirm :+1:

front-end/src/components/LoginForm.jsx
jmmxp: Good question, I honestly don't know what is the usual practice is in JS/TS, but I usually try to stay away from having too many levels of indentation.

ChrisChrisLoLo: https://www.freecodecamp.org/news/how-to-deal-with-nested-callbacks-and-avoid-callback-hell-1bc8dc4a2012/
Yeah nesting callbacks definitely isn't great after a while, you bring up a good point :+1: . I'll have to chew some stuff over in my head to figure out what the nicest way of doing promises are. In the meantime I'll strive towards `.then` chaining

front-end/src/App.tsx
jmmxp: Cleared this up with Chris offline, had some confusion from the `.filter` line above.

front-end/src/components/LoginForm.jsx
jmmxp: Sounds good!

jmmxp:
We could probably get nodes upon opening the page as well, but whenever we use a Node's credentials to make a request, we *are* sending Basic Auth credentials, so it implies that it's an authenticated request anyways. (i.e., if someone local on the other person's server makes the same request they should probably be authed/logged in as well).

TLDR; sending credentials implies we should be logged in

we might have to discuss if this really kills the functionality of the site for users that aren't logged in, but I wouldn't be too worried about that

PR #161, opened 2021-03-24
Likes

ChrisChrisLoLo:
Add ability to like a Post

Remove some logging

Screenshot_20210324_173448

Screenshot_20210324_173512


front-end/src/components/LikeListItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+interface Props {
+  like: Like;
+}
+
+/**
+ * Author list component to show list of authors
+ * @param props
+ */
+export default function LikelListItem(props: Props) {
```
jmmxp: nit; Is the naming of this a typo? `LikelListItem` -> `LikeListItem`?

front-end/src/components/LikesFeed.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * Shows all the likes recieved
+ * @param props
+ */
+export default function LikesFeed(props: Props) {
+  return (
+    <div>
+      <h3>Likes</h3>
+      {props.likesList?.length !== 0 ?
+       props.likesList?.map((like: Like, i) => <LikelListItem like={like} key={i}/>):
+       <Card body className="text-center"><CardBody><CardTitle tag="h5" >No likes found. Everyone hates you :(</CardTitle></CardBody></Card>
```
jmmxp, line 19: :(

front-end/src/components/PostListItem.tsx
```diff
@@ -25,14 +28,52 @@ export default function PostListItem(props:Props) {
...
+                  like
+                ).then( res => {
+                  alert('You liked the post!');
+                  setHasLiked(true);
+                });
+              }
+            )
+    }else{
+      console.error("User is not logged in, cannot like!");
+    }
```
jmmxp: nit; could we avoid nesting Promises and follow what I did in refactoring Login/Register [here](https://github.com/CMPUT404W21T02/group-project/blob/872705ea17bb9f964f03f16ad0f9906bc1751339/front-end/src/components/LoginForm.jsx) to make it more readable?

wouldn't mind it if you kept it this way though, the code I changed was much more messy because of the multiple `.catch` statements.

jmmxp (commented):
Just some nitpicks 

Also, is that "Liked" button clickable on a Post that the Author has already liked? If so we might want to just hide that "Liked" text or make it not blue etc.
(don't need to change it in this PR but good to keep in mind for the future).

front-end/src/components/PostListItem.tsx
```diff
@@ -25,14 +28,52 @@ export default function PostListItem(props:Props) {
...
+                }
+                axios.post(`${process.env.REACT_APP_API_URL}/api/author/${props.post.author.id}/inbox/`,
+                  like
+                ).then( res => {
+                  alert('You liked the post!');
+                  setHasLiked(true);
+                });
+              }
+            )
+    }else{
```
jmmxp: Another nit; but maybe just do the below to avoid unnecessary nesting / indentation?
```javascript
if (!props.loggedInUser) {
  console.error(...)  
  return;
}

// make requests here
```

front-end/src/components/LikeListItem.tsx
ChrisChrisLoLo: Good catch :+1:

jmmxp (approved):
LGTM just nits

front-end/src/components/PostListItem.tsx
ChrisChrisLoLo: I agree, added change :+1:

front-end/src/components/PostListItem.tsx
ChrisChrisLoLo: Fair point, added change

PR #165, opened 2021-03-25
Create Axios wrapper

jmmxp:
Created an Axios wrapper that should replace `axios` to make API requests whenever we use an endpoint that could be cross-server.

TODO: Still needs to be actually constructed and used instead of axios.

Closing PR in favour of #167

### DIFF
--- a/front-end/src/App.tsx
+++ b/front-end/src/App.tsx
@@ -50,9 +50,9 @@ function App() {
 
       // Keep track of a Node representing our own server
       setNodes([...nodes, {
-        "host": host,
-        "username": loggedInUser?.username, 
-        "password": loggedInUser?.password
+        host: host,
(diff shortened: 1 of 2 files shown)


front-end/src/App.tsx
```diff
@@ -50,9 +50,9 @@ function App() {
 
       // Keep track of a Node representing our own server
       setNodes([...nodes, {
-        "host": host,
-        "username": loggedInUser?.username, 
-        "password": loggedInUser?.password
+        host: host,
+        username: loggedInUser?.username, 
+        password: loggedInUser?.password
```
jmmxp, line 55: Just fixing up some old code - was too used to Python's definition of dictionaries and didn't realize you could define keys without quotes.

PR #166, opened 2021-03-25
Setup sending created posts to followers inbox

VinnyLuu:
When a post is created, the front end will now get all the followers of the user and send the post to each of their inboxes. Also setup a script to run the frontend locally without the  REACT_APP_API_URL environment variable.

user abc4's inbox shows abc3's post, since abc4 is following abc3
image

Also fixed followbutton urls to have the trailing slash and use format strings

### DIFF
--- a/front-end/src/components/CreateEditPostModal.tsx
+++ b/front-end/src/components/CreateEditPostModal.tsx
@@ -3,6 +3,7 @@ import React, { useState } from "react";
 import { Alert, Form, FormGroup, Input, Label, Modal, ModalBody, ModalFooter, ModalHeader } from "reactstrap";
 import { Post, PostContent, PostContentType, PostVisibility } from "../types/Post";
 import { UserLogin } from "../types/UserLogin";
+import { Author } from "../types/Author";
 import PostContentEl from "./PostContentEl";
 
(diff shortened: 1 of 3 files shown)


front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -103,8 +104,17 @@ export default function CreateEditPostModal(props: Props){
...
+            handleRes(post)
+            return post
+          }).then(post => {
+            // send this post to all followers
+            axios.get(`${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/followers/`).then(res => {
+              let followingList: Author[] = res.data.items;
+              followingList.forEach(follower => {
+                axios.post(`${process.env.REACT_APP_API_URL}/api/author/${follower.id}/inbox/`, post.data);
+              });
+            });
```
jmmxp, line 117: On second thought, you can just get rid of this 2nd .then and put this code in place of "return post" above

jmmxp (approved):
LGTM

ChrisChrisLoLo (approved):
LGTM, good stuff :shipit:

PR #167, opened 2021-03-25
Create AxiosWrapper, refactor all axios requests to use the wrapper

jmmxp:
AxiosWrapper keeps a static array of Nodes, which gets set whenever Nodes is updated. 

Whenever AxiosWrapper makes requests, we check the URL of the request, and see if the URL contains any of the Nodes' hostnames. If it does, we use those credentials stored in that Node to make the request.

If no credentials exist, we make a normal axios request (with no credentials).


front-end/src/App.tsx
```diff
@@ -50,9 +51,9 @@ function App() {
 
       // Keep track of a Node representing our own server
       setNodes([...nodes, {
-        "host": host,
-        "username": loggedInUser?.username, 
-        "password": loggedInUser?.password
+        host: host,
+        username: loggedInUser?.username, 
+        password: loggedInUser?.password
```
jmmxp, line 56: Just refactoring this to create keys without using quotes, was too used to Python dictionaries.

front-end/tsconfig.json
```diff
@@ -18,7 +18,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "noImplicitAny": false
```
ChrisChrisLoLo: Is there a way we can go about not disabling this? I think it could open up a few future bugs considering the number of devs we currently have. These bugs could either make our app less stable or harder to debug, since I don't know if we can catch all of them in the PRs

front-end/src/helpers/AxiosWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+export class AxiosWrapper {
+  static nodes: Node[] = []
+  
+  /**
+   * Given a URL, checks the host and retrieves the credentials requests to that URL should be made with.
+   * @param url to get the credentials for
+   */
+  private static getCredentialsForUrl(url: string) {
+    let nodes = AxiosWrapper.nodes.filter(n => url.includes(n.host));
+
```
ChrisChrisLoLo, line 13: Could we log an error if no such credential was found for a given host?

ChrisChrisLoLo (changes requested):
Thanks for getting this done in a short amount of time!
I think the implementation of this wrapper is pretty clever

My only concern with this PR is the widening of TS rules. I think given the diverse technical background of our team, we want TS to be as strict as possible to prevent any type of coercion bugs or undefined exceptions. I think the cost of debugging/ potential demo bugs could outweigh the problem we currently have in this PR, though I could be wrong.

Could you elaborate on what situation is a pain point currently? We can discuss if it can be resolved without changing the TS config. Worst comes to worst, we can change the TS config if need be

front-end/src/App.tsx
```diff
@@ -63,6 +64,7 @@ function App() {
     if (nodes.length === 0){
       localStorage.removeItem(LOCAL_STORAGE_NODES);
```
jmmxp, line 66: TODO: `AxiosWrapper.nodes = []`, so that the wrapper loses its credentials when a logout happens.

front-end/src/helpers/AxiosWrapper.tsx
jmmxp: Done

front-end/tsconfig.json
jmmxp: Talked with Chris offline, simply changing .then and .catch callback parameters to (res: any) and (err: any) prevents me from having to disable this TS config lint. Changes are now in.

front-end/src/App.tsx
jmmxp: Done.

front-end/src/pages/SettingsPage.tsx
```diff
@@ -35,9 +35,9 @@ const SettingsPage = (props: Props) => {
 
     // get current author data
     function getAuthorData() {
-        AxiosWrapper.get(authorUrl).then(res => {
+        AxiosWrapper.get(authorUrl).then((res: any) => {
             setUnchangedData(res.data);
-        }).catch(err => {
+        }).catch((err: any) => {
```
ChrisChrisLoLo, line 40: If anyone is curious about what can be done to type this down the road, this might be useful: https://stackoverflow.com/questions/62217642/react-and-typescript-which-types-for-axios-response

The current implementation is good as is as it localizes `any` to a few method calls, and we didn't really use exhastive typing on `res` or `err` to begin with :+1:

ChrisChrisLoLo (approved):
LGTM! :shipit:

front-end/src/pages/SettingsPage.tsx
jmmxp: Thanks for this! This clears things up.

PR #172, opened 2021-03-25
GET authors from all possible servers when searching

jmmxp:
I loop through the array of Nodes we have and make a request to each of their /api/authors/ endpoints, and collect the result using `Promises.allSettled`.

I then get the authors out of all Promises that resolved successfully (fulfilled) and display that.
Other endpoints we're making requests to still need to be generalized for across cross-server (follow, inbox/like)


VinnyLuu (approved):
LGTM

PR #174, opened 2021-03-26
Change request URLs to support cross-server actions

jmmxp:
Instead of always prepending our API endpoint environment variable to request URLs, we look at Author objects' host fields and prepend that to our requests.

This is now being done for: sending likes to Inbox, sending posts to Inbox, following an Author, and checking if we are already following an Author.

Note that our environment variable has no trailing slash, but `author.host` does have trailing slashes, so there is a bit of mismatch in the URLs we make requests to: author.host + "api/author" as opposed to envvar + "/api/author"


VinnyLuu (approved):
LGTM

PR #175, opened 2021-03-26
Private post are now only sent to friends inboxes

VinnyLuu:
Private post are now sent only to friends by GETing all the friends and posting to each of their inboxes.

Author abc3 is following abc2 but abc2 is not following back. Therefore abc3's inbox is empty when abc2 made a private post.
image

Author abc1 is following abc2 and vice versa, therefore they are friends. abc1 now sees the private post that abc2 made.
image

### DIFF
--- a/front-end/src/components/CreateEditPostModal.tsx
+++ b/front-end/src/components/CreateEditPostModal.tsx
@@ -103,13 +103,23 @@ export default function CreateEditPostModal(props: Props){
             handleRes(post)
             return post
           }).then((post: any) => {
-            // send this post to all followers
-            AxiosWrapper.get(`${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/followers/`).then((res: any) => {
-              let followingList: Author[] = res.data.items;
-              followingList.forEach(follower => {
-                AxiosWrapper.post(`${process.env.REACT_APP_API_URL}/api/author/${follower.id}/inbox/`, post.data);
+            if (visibility === PostVisibility.FRIENDS) {
+              // send this post to friends only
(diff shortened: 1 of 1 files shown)


front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -103,13 +103,23 @@ export default function CreateEditPostModal(props: Props){
...
+                  AxiosWrapper.post(`${process.env.REACT_APP_API_URL}/api/author/${friend.id}/inbox/`, post.data);
+                });
+              })
+            } else {
+              // send this post to all followers (which are friends and followers)
+              AxiosWrapper.get(`${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/followers/`).then((res: any) => {
+                let followingList: Author[] = res.data.items;
+                followingList.forEach(follower => {
+                  AxiosWrapper.post(`${process.env.REACT_APP_API_URL}/api/author/${follower.id}/inbox/`, post.data);
+                });
```
jmmxp: What about this (for cleaner code)?

```
if visibility is PostVisibility.FRIENDS{
  return AxiosWrapper.get(...)
} else {
  return AxiosWrapper.get(...)
}.then(authors => {
  // post to their inboxes.
})
```

jmmxp (approved):
LGTM but style/clean code comment

front-end/src/components/CreateEditPostModal.tsx
VinnyLuu: With that way, I'll need to return both post and the list to the next promise to post to the inboxes. Keeping it this way I think would be cleaner.

PR #176, opened 2021-03-26
Change URL for getting Post likes (to work cross-server)

jmmxp:
Now if the post is from a server other than ours, it should properly display the likes.


VinnyLuu (approved):
LGTM

PR #177, opened 2021-03-26
Auth endpoints that we use cross-server

jmmxp:
To properly ensure that these endpoints only work when authenticated, I've added that they require BasicAuth in the back-end.

Note that we're trying to target:

GET all authors
POST like/ POST post to an inbox
GET/POST follower for an author
POST comments

Note that we can only put auth classes on a whole ViewSet and not specific methods of a ViewSet. If we can figure out how to do it on specific methods of a ViewSet, let me know. (All methods endpoint provides will require auth because of this).


VinnyLuu (approved):
LGTM

PR #178, opened 2021-03-26
Add trailing slash to inbox URL

jmmxp:
Right now when axios makes requests, it sees that we get a 301 response and Google Chrome makes a request to that new URL. This might be causing some issues with fetching the inbox.


VinnyLuu (approved):
LGTM

PR #179, opened 2021-03-26
Unauth inbox GET

jmmxp:
Doing this to fix a bug where, our front-end tries to make requests to the inbox for the FriendsList before it has the proper Nodes (credentials) to. We will need to fix that by either:

1. forcing it to make a new request whenever Nodes changes OR
2. moving friend requests over to the Inbox (which makes a fresh request whenever you click it).


VinnyLuu (approved):
LGTM

PR #180, opened 2021-03-26
Disable prepend if the post is not Public

jmmxp:
We shouldn't be prepending the post to the feed if the post is not Public. (private posts are sent out to the inboxes of others and should not be visible on the front feed, it is only visible on the Profile feed).

### DIFF
--- a/front-end/src/components/CreateEditPostModal.tsx
+++ b/front-end/src/components/CreateEditPostModal.tsx
@@ -141,7 +141,7 @@ export default function CreateEditPostModal(props: Props){
     } else if (res.status >= 200) {
       const post:Post = res.data;
       if(isCreate){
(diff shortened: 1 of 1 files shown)


front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -141,7 +141,7 @@ export default function CreateEditPostModal(props: Props){
     } else if (res.status >= 200) {
       const post:Post = res.data;
       if(isCreate){
-        if(props.prependToFeed !== undefined){
+        if(props.prependToFeed !== undefined && post.visibility === "Public"){
```
VinnyLuu: you can use the static variable instead of "Public". i think its Post.visibility.Public

VinnyLuu (approved):
LGTM

PR #181, opened 2021-03-27
Fix links to Followers and Following list on profile page

VinnyLuu:
Using CardLinks with href to go to the followers and following lists are broken on our netlify instances. I think the issue was that with href, its assuming that it is a path and follows the path /author/authorid/followers and looks for a followers file. Wrapping the button with a Link instead should work since it is using react dom router to change the page. The button do not look any different and actually work better since now you can click anywhere on the button to redirect, instead of actually on the text in the button.

image


raysarinas (approved):
LGTM, just be sure to test out the tabs and that the URL changes as it should

PR #182, opened 2021-03-27
Update wrapper to request for nodes every time it makes a cross-server request

jmmxp:
The AxiosWrapper now makes a call to `/api/nodes` every single time it gets a URL that requires a cross-server request. If it sees that the URL is local (our own server), then it avoids making this extra /nodes call, and sends the credentials we have for the current UserLogin.

This will prevent us from having to keep track of the nodes client-side, which can run into a lot of race conditions with localStorage. Warning that we still have race conditions with UserLogin being used before it is read from localStorage, but this should be easier to handle than not having the proper Nodes.

### DIFF
--- a/front-end/src/pages/AuthorPage.tsx
+++ b/front-end/src/pages/AuthorPage.tsx
@@ -48,7 +48,7 @@ export default function AuthorPage(props: any) {
     if (props.loggedInUser) {
       // get whether user is follower of author IF not looking at our own profile
       if (!props.location.pathname.includes(props.loggedInUser.authorId)) {
-        AxiosWrapper.get(authorUrl + "/followers/" + props.loggedInUser.authorId, props.loggedInUser).then((res: any) => {
+        AxiosWrapper.get(`${authorUrl}/followers/${props.loggedInUser.authorId}/`, props.loggedInUser).then((res: any) => {
           setIsFollower(true);
         }).catch((err: any) => {
           // 404 is not a follower
--- a/front-end/src/pages/AuthorPage.tsx
+++ b/front-end/src/pages/AuthorPage.tsx
@@ -36,7 +36,7 @@ export default function AuthorPage(props: any) {
 
   // After clicking the profile navlink, get the appropriate author info and data
   useEffect(() => {
-    AxiosWrapper.get(authorUrl).then((res: any) => {
+    AxiosWrapper.get(authorUrl, props.loggedInUser).then((res: any) => {
       const authorOb: Author = res.data;
       setAuthor(authorOb);
       setResponseMessage(200);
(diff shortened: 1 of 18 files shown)


front-end/src/pages/AuthorPage.tsx
```diff
@@ -48,15 +48,15 @@ export default function AuthorPage(props: any) {
     if (props.loggedInUser) {
       // get whether user is follower of author IF not looking at our own profile
       if (!props.location.pathname.includes(props.loggedInUser.authorId)) {
-        AxiosWrapper.get(authorUrl + "/followers/" + props.loggedInUser.authorId).then((res: any) => {
+        AxiosWrapper.get(authorUrl + "/followers/" + props.loggedInUser.authorId, props.loggedInUser).then((res: any) => {
```
VinnyLuu: can you add this to have a trailing slash?

front-end/src/pages/PostDetailPage.tsx
```diff
@@ -28,10 +29,10 @@ export default function PostDetailPage(props: Props) {
   useEffect(() => {
     let getPromise;
     if (props.loggedInUser) {
-      getPromise = AxiosWrapper.get(`${process.env.REACT_APP_API_URL}/api/posts/${postId}`);
+      getPromise = AxiosWrapper.get(`${process.env.REACT_APP_API_URL}/api/posts/${postId}`, props.loggedInUser);
```
VinnyLuu: trailing slash

front-end/src/pages/PostDetailPage.tsx
```diff
@@ -28,10 +29,10 @@ export default function PostDetailPage(props: Props) {
...
     }
     else {
-      getPromise = AxiosWrapper.get(`${process.env.REACT_APP_API_URL}/api/posts/${props.location.pathname}`,);
+      getPromise = axios.get(`${process.env.REACT_APP_API_URL}/api/posts/${props.location.pathname}`);
```
VinnyLuu: trailing slash. This isn't using the AxiosWrapper since props.loggedInUser is undefined right?

VinnyLuu (approved):
LGTM, good stuff boss

front-end/src/pages/PostDetailPage.tsx
jmmxp: Yeah, that was the original reason I had this here, but I later changed the wrapper to properly handle undefined. I just refactored this to return to using the wrapper now.

jmmxp: Not changing trailing slash since it's not needed here

front-end/src/pages/PostDetailPage.tsx
jmmxp: Changed

front-end/src/pages/AuthorPage.tsx
jmmxp: Changed

PR #183, opened 2021-03-27
Reverse inbox posts so that they are in order

jmmxp:
Currently, posts POSTed to someones inbox are added to the bottom of the inbox. This results in the posts on the Inbox page being shown as oldest first rather than newest first. It is also much harder to sort these posts because they are stored as JSONFields in our backend.

Reversing the list should work for the foreseeable future, the main difference with this post list and our public feed is that the inbox feed is read-only.


VinnyLuu (approved):
LGTM

PR #185, opened 2021-03-27
Removed authentication on authors list endpoint

VinnyLuu:
Currently Amy's group did not expect us to authenticate the authors list endpoint. The plan is to remove authentication on this endpoint until their demo is over and it is safe to put authentication back in.


jmmxp (approved):
LGTM

PR #186, opened 2021-03-27
Update author search to expect trailing slash from nodes.

jmmxp:
Currently "host" seems to mean with the trailing slash (Author JSON in the requirements spec includes the trailing slash).

I've changed this request to expect that, meaning that in the back-end, we should save Node hosts with the trailing slash. This helps us stay consistent with other cross server requests.


VinnyLuu (approved):
LGTM

PR #187, opened 2021-03-27
Changed PUT endpoint for follower to match other teams input.

VinnyLuu:
One of our groups that we are working, when PUTing to the follow endpoint, they are sending a friend request json which contains both the sender and the reciever and a type follow. This is different from what we were doing originally, which was just expecting the author object of the sender.


jmmxp (approved):
LGTM

PR #193, opened 2021-03-28
Modify comments back-end to match what we want

jmmxp:
Removed contentType field from comments, changed author to be a JSONField, and changed postId to be a Foreign Key to a post. Updated tests to match the above changes.

Note that I had to nuke migrations since we completely overhauled the 'post' field. Will need to nuke migrations on our deployed Heroku instance as well.


back-end/api/quickstart/views.py
```diff
@@ -162,13 +162,19 @@ class CommentViewSet(viewsets.ModelViewSet):
     lookup_field = 'id'
 
     def retrieve(self, request, author, post):
+        queryset = Comment.objects.filter(post=post)
```
VinnyLuu, line 165: Shouldn't this post be the actual post object not the id? since its a foreign key

back-end/api/quickstart/models.py
```diff
@@ -48,11 +48,10 @@ class Visibility(models.TextChoices):
 # An Author can communicate with other Authors by commenting on posts. 
 class Comment(models.Model):
   id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+  post = models.ForeignKey(Post, on_delete=models.CASCADE)
   type = 'comment'
-  postId = models.TextField()
-  author = models.TextField()
+  author = models.JSONField()
   comment = models.TextField()
-  contentType = models.TextField()
```
VinnyLuu, line 55: Is it alright to remove this? Are comments just gonna be plain text?

jmmxp: Markdown support for comments doesn't seem to be in the requirements, so it seems that we don't need this.

VinnyLuu: Awesome sounds good!

VinnyLuu (approved):
LGTM

back-end/api/quickstart/views.py
jmmxp: This seems to work (Django is smart enough to take in the ID).

PR #195, opened 2021-03-28
Add comments UI

raysarinas:
We can leave comments on posts now.

A few things to note:
- Made the post details modal wider/fatter 
- We can only view comments if we're logged in (need credentials to GET them)
- We can't "like" comments atm. This is something I think we can work on and do independently
- Not sure what to do about pagination or how we want to handle that - probably something to discuss
- ~~After we make a comment, the list doesn't update/refresh~~ done

Post details modal looks like this now:
Screen Shot 2021-03-28 at 4 27 17 PM

What it looks like when you aren't logged in:
Screen Shot 2021-03-28 at 4 37 16 PM


front-end/src/components/CommentFormElement.tsx
```diff
@@ -0,0 +1,85 @@
...
+      // Successfully commented on post, so clear the text box
+      setCommentContent("");
+      // TODO: append the new comment to the list?
+    } else if (res.status >= 400) {
+      // Error in commenting on post
+      setShowError(true)
+    }
+  }
+
+  if (!props.loggedInUser) {
```
ChrisChrisLoLo: Nit/my suggestion could be unorthodox: it might be tidier if we brought this conditional to the near top of the function kind of like the following:
```ts
function component(){
   if(!props.loggedInUser){
       return jsxPromptToLogIn
   } else {
       // Run logic
       return jsxCardShowingCommentsForm
   }
}
```
This way I can quickly tell that if `props.loggedInUser` is `undefined`, that something will be returned, and there won't be any edge cases if it's undefined. It's a nitpick so it's something that could be done in the future

front-end/src/components/CommentFormElement.tsx
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <Card body className="text-center">
+        <CardTitle>Log in to make and see comments!</CardTitle>
```
ChrisChrisLoLo: Is it possible to make it so that we can see comments if we're not logged in? Maybe I missed the discussion where this api route needs to be authenticated to work

front-end/src/components/CommentList.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  const [commentList, setCommentList] = useState<PostComment[] | undefined>(undefined);
+
+  useEffect(() => {
+    AxiosWrapper.get(`${props.postAuthor.url}posts/${props.postId}/comments/`, props.loggedInUser).then((res: any) => {
+      // TODO: filters for friend posts?
+      console.log(res.data);
+      const comments: PostComment[] = res.data;
+      // Reverse the posts so that they are in order (from newest to oldest).
+      setCommentList(comments.reverse());
```
ChrisChrisLoLo, line 33: This is probably good! It might get scrambled, as I had this problem when sorting the posts with this method. I don't think the consequences of this getting scrambled are a big deal, but if it is, we might want the back-end to handle the sorting

front-end/src/components/CommentFormElement.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Delete Post modal, confirmation to delete post as well as removing from list on success
+ * @param props 
+ */
+export default function CommentFormElement(props: Props) {
+
+  const [showError, setShowError] = useState(false);
+  const [commentContent, setCommentContent] = useState("");
+  const [commenter, setCommenter] = useState<Author | undefined>(undefined);
```
ChrisChrisLoLo: This is a note more so for conversation. I personally don't know how to get around not using `undefined`. I used `useState<Author | undefined>` in this project myself, and I think the codebase has become a bit more complicated as a result. I wonder if it's possible to use something like `useState<Author | Promise<Author>>` or similar to prevent working with `undefined` values.

If you have any suggestions about how to go about this let me know :+1:. It's something I'd like to work around when I'm writing this kind of code

front-end/src/components/CommentList.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }).catch((err: any) => {
+      console.error(err);
+    });
+  }, []);
+
+  return (
+    <>
+      <Col>
```
ChrisChrisLoLo, line 50: nit: It's a bit less idiomatic if we use `<Cols>` without `<Containers>` and `<Rows>`. e.g.
```tsx
<Container>
  <Col>
    <Row>
    </Row>
  </Col>
</Container>
```
I've found from past experience that using components in ways that they weren't exactly intended occasionally led to undesired css positioning and media queries that were hard to debug. I also find it a bit harder to read if components are used in places outside of their usual context. I think it should be fine in this instance, but I think it would be good to use Reactstrap the way the framework intends to ensure that the CSS behaves consistently going forward

ChrisChrisLoLo (approved):
LGTM :shipit: 

I think you have the TS and React patterns really down at this point, at least by my expectations. Also really appreciate the clear notes about what the feature can and cannot do.

Most comments are fairly nitpicky. Aside from the comment about Reactstrap, I think it's fine if the PR gets merged in as-is. 

In terms of not getting the comments to update, feel free to check out https://github.com/CMPUT404W21T02/group-project/pull/108/files as a reference for how the posts do it if you have time. If you don't have time we can create a new ticket in the backlog for it.

front-end/src/components/CommentFormElement.tsx
ChrisChrisLoLo: Your PR answers this. Resolving

front-end/src/components/CommentList.tsx
raysarinas: agreed, sounds like something to keep in mind then. I just followed the way we're currently sorting inbox posts/entries on the homepage

front-end/src/components/CommentList.tsx
raysarinas: put the `<Col>` inside a `<Container>`. Wrapping the cards/comment items in a `<Row>` made the components behave weird

front-end/src/components/CommentFormElement.tsx
raysarinas: done

PR #196, opened 2021-03-30
Reshare

Monica-Uwalaka:
Can rehare public post in private feed now.


.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "python.pythonPath": "back-end/venv/bin/python3"
+  "python.pythonPath": "C:\\Python32\\python.exe"
```
VinnyLuu: This is for your own computer right? Should probably change this back to what it was.

front-end/src/components/PostListItem.tsx
```diff
@@ -90,6 +125,9 @@ export default function PostListItem(props: Props) {
           {EditCardLink()}
           {DeleteCardLink()}
           {LikeCardLink()}
+          {props.isReshareable ? (ReshareCardLink()): null}
+          
+
```
VinnyLuu: nit: can remove all these newlines

front-end/src/components/PostListItem.tsx
```diff
@@ -90,6 +125,9 @@ export default function PostListItem(props: Props) {
           {EditCardLink()}
           {DeleteCardLink()}
           {LikeCardLink()}
+          {props.isReshareable ? (ReshareCardLink()): null}
```
VinnyLuu: I think you can move the condition to the ReshareCardLink function and it should still work, just to keep it consistent with how these other ones are doing it.

front-end/src/components/PostListItem.tsx
```diff
@@ -66,6 +70,33 @@ export default function PostListItem(props: Props) {
...
+        contentType: post.contentType,
+        content: post.content,
+        categories: post.categories
+      }
+      AxiosWrapper.post(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser?.authorId + "/posts/", data, props.loggedInUser)
+        .then((res: any) => {
+            if (res.status >= 400) {
+                setShowError(true)
+            } else if (res.status === 201) {
+                //props.history.push("/")
```
VinnyLuu: We are taking care of sending the posts to followers inboxes on the frontend, not the backend so you haven't shared the posts to followers inboxes yet. Here with res.data is the post with it you should send to followers inboxes. You can take a look at how its done by going to the CreateAndEditPostModal file. In there, you can take that code and put it here and it should work. We can maybe discuss later about if we want to be able to customize whether or not its sent to friends vs followers but for now, I think we can just keep it to all followers for the deadline.

VinnyLuu (changes requested):

.vscode/settings.json
Monica-Uwalaka: Yes I think so, I'll change it

front-end/src/components/PostListItem.tsx
```diff
@@ -66,6 +70,37 @@ export default function PostListItem(props: Props) {
...
+
+            return AxiosWrapper.get(authorsUrl, props.loggedInUser);
+          }).then((res: any) => {
+            let authors: Author[] = res.data.items;
+            authors.forEach(a => {
+              AxiosWrapper.post(`${a.host}api/author/${a.id}/inbox/`, post, props.loggedInUser);
+            });
+          }).catch((error: any) => {
+            setShowError(true)
+          })
```
VinnyLuu: nit. extra indentation

front-end/src/components/PostListItem.tsx
```diff
@@ -90,6 +129,8 @@ export default function PostListItem(props: Props) {
           {EditCardLink()}
           {DeleteCardLink()}
           {LikeCardLink()}
+          {ReshareCardLink()}
+          {/* {props.isReshareable ? (ReshareCardLink()): null} */}
```
VinnyLuu: nit: you can delete this comment

.vscode/settings.json
VinnyLuu: Is there still another commit for this change?

front-end/src/components/PostListItem.tsx
```diff
@@ -75,6 +110,10 @@ export default function PostListItem(props: Props) {
       : <CardLink onClick={() => likePost()}>Like</CardLink>
     : null;
 
+  //reshare cardlink 
+  const ReshareCardLink = () => props.loggedInUser !== undefined && isAuthorPost && props.isReshareable ?<CardLink onClick={() => reshare(props.post)}>share</CardLink> : null;
```
VinnyLuu: Should not have isAuthorPost in this condition. We want to share posts made by other authors as well.

front-end/src/pages/HomePage.tsx
```diff
@@ -117,7 +117,7 @@ export default function HomePage(props: any) {
                   <Row>
                     <Col></Col>
                     <Col sm={8}>
-                      {<PostList postEntries={postEntries} setPostEntries={setPostEntries} loggedInUser={props.loggedInUser} />}
+                      {<PostList postEntries={postEntries} setPostEntries={setPostEntries} loggedInUser={props.loggedInUser} isResharable={false}/>}
```
VinnyLuu: This is for public posts correct? I think it we want it to be that all posts are shareable. To make this simpler I think that every post can be sharable. This way you can remove this boolean and remove the conditions with this bool. If the TA says, otherwise, it wouldn't be hard to put this bool back in.

VinnyLuu (changes requested):

jmmxp:
Since this is a PR with UI changes, could you provide a screenshot of what the reshare button looks like?

front-end/src/components/PostListItem.tsx
jmmxp: This should actually be !isAuthorPost right? We should only be able to reshare posts that aren't ours

VinnyLuu: Ah right good point!

front-end/src/components/PostListItem.tsx
Monica-Uwalaka: the whole block?

VinnyLuu: yea from lines 86 to 100

VinnyLuu: its just that the indent is 4 spaces instead of 2

front-end/src/pages/HomePage.tsx
Monica-Uwalaka: I'll leave everything as true for now. So if we have to change it for any reason we only have to change the Boolean values.

Monica-Uwalaka:
Screenshot (158)
Screenshot (157)

front-end/src/components/PostListItem.tsx
```diff
@@ -111,7 +111,7 @@ export default function PostListItem(props: Props) {
     : null;
 
   //reshare cardlink 
-  const ReshareCardLink = () => props.loggedInUser !== undefined && isAuthorPost && props.isReshareable ?<CardLink onClick={() => reshare(props.post)}>share</CardLink> : null;
```
VinnyLuu: Sorry as what @Jmmxp pointed out it should be !isAuthorPost instead, so that posts that aren't you're own are reshareable and not what I said earlier.

front-end/src/components/PostListItem.tsx
```diff
@@ -75,6 +110,10 @@ export default function PostListItem(props: Props) {
       : <CardLink onClick={() => likePost()}>Like</CardLink>
     : null;
 
+  //reshare cardlink 
+  const ReshareCardLink = () => props.loggedInUser !== undefined && props.isReshareable ?<CardLink onClick={() => reshare(props.post)}>share</CardLink> : null;
```
VinnyLuu: Could you capitalize the share button

VinnyLuu (approved):
LGTM after you capitalize the share button 👍

PR #199, opened 2021-03-31
Created helper class to check response codes

VinnyLuu:
Other servers have different response codes than what we would expect from our own server. Created a helper class to tell if it was a success or not, instead of expecting a specific status code.

### DIFF
--- a/front-end/src/helpers/ResponseHelper.tsx
+++ b/front-end/src/helpers/ResponseHelper.tsx
@@ -1,0 +1,16 @@
+import { AxiosResponse } from "axios";
+
+// Helpers to check category of response code
+export class ResponseHelper {
+  static isSuccess = (res:AxiosResponse) => {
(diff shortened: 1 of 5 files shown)


front-end/src/helpers/ResponseHelper.tsx
```diff
@@ -0,0 +1,16 @@
+import { AxiosResponse } from "axios";
+
+// Helpers to check category of response code
+export class ResponseHelper {
+  static isSuccess = (res:AxiosResponse) => {
+    return res.status >= 200 && res.status < 300;
+  }
+
+  static isRequestError = (res:AxiosResponse) => {
+    return res.status >= 400 && res.status < 500;
```
jmmxp: nit isClientError not isRequestError. request error is specifically 400 bad request.

jmmxp (approved):
LGTM

PR #203, opened 2021-04-01
GitHub

ChrisChrisLoLo:
Add github integration. Takes the github url and uses it to get all events. We can tidy it up a bit later.

Made some light refactorings here and there as well

Screenshot_20210401_152741

### DIFF
--- a/front-end/src/components/GithubFeed.tsx
+++ b/front-end/src/components/GithubFeed.tsx
@@ -19,8 +19,8 @@ export default function GithubFeed(props: Props) {
 
   useEffect(() => {
     const usernameMatches = props.github.match(/https?:\/\/github.com\/(.+)/);
-    console.log(usernameMatches, usernameMatches?.length !== undefined, usernameMatches?.length === 2);
-    if (usernameMatches?.length !== undefined && usernameMatches.length === 2){
+    // console.log(usernameMatches, usernameMatches?.length !== undefined, usernameMatches?.length === 2);
+    if (usernameMatches && usernameMatches.length === 2){
       axios.get(`https://api.github.com/users/${usernameMatches[1]}/events`).then((res) => {
         setGithubEvents(res.data)
       });
--- a/front-end/src/components/GithubFeed.tsx
+++ b/front-end/src/components/GithubFeed.tsx
@@ -1,0 +1,45 @@
+import axios from "axios";
+import React, { useEffect, useState } from "react";
+import { GithubEvent } from "../types/Github";
+import GithubFeedItem from "./GithubFeedItem";
+
+interface Props {
+  // github URL
+  github: string;
+}
+
+/**
+ * Author list component to show list of authors
+ * @param props
+ */
+export default function GithubFeed(props: Props) {
+  let [githubEvents, setGithubEvents] = useState<GithubEvent[]>([]);
+  let [error, setError] = useState<string>("");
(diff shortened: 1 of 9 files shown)


front-end/src/components/GithubFeed.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+export default function GithubFeed(props: Props) {
+  let [githubEvents, setGithubEvents] = useState<GithubEvent[]>([]);
+  let [error, setError] = useState<string>("");
+  let [loading, setLoading] = useState<boolean>(true);
+
+  useEffect(() => {
+    const usernameMatches = props.github.match(/https?:\/\/github.com\/(.+)/);
+    console.log(usernameMatches, usernameMatches?.length !== undefined, usernameMatches?.length === 2);
+    if (usernameMatches?.length !== undefined && usernameMatches.length === 2){
```
raysarinas: I think this `if` statement can be simplified i.e. `if (usernameMatches && usernameMatches.length === 2)`

front-end/src/components/GithubFeed.tsx
```diff
@@ -0,0 +1,45 @@
...
+      axios.get(`https://api.github.com/users/${usernameMatches[1]}/events`).then((res) => {
+        setGithubEvents(res.data)
+      });
+    } else {
+      setError("The github url provided isn't valid :(");
```
raysarinas, line 28: Might be a good idea to use Vinny's response code helper class - or maybe extend it so that it can help with consolidating similar errors

front-end/src/pages/AuthorPage.tsx
```diff
@@ -123,7 +127,14 @@ export default function AuthorPage(props: any) {
...
         <Col>
-          {author && displayPosts()}
+          {author && props.loggedInUser && <PostList postEntries={postEntries} setPostEntries={setPostEntries} loggedInUser={props.loggedInUser} isResharable={true}/>}
+          
+          {author && isValidGithub(author.github) && 
+            <>
+              <h3>Github feed</h3>
+              <GithubFeed github={author.github}/>
+            </>
+          }
```
raysarinas, line 137: If we're going with not showing author posts on their profile, we could have it so that our profile page just shows our GitHub feed since it's the only GitHub activity that we're able to see (I think). Otherwise, we should talk about where to put the GitHub feed. Could have another tab or navlink to our GitHub feed like we have the public and inbox feeds

raysarinas (approved):
LGTM. just a few things I commented that might be worth taking a look at, but I think should be good to merge in!

front-end/src/components/GithubFeed.tsx
ChrisChrisLoLo: Good catch

front-end/src/pages/AuthorPage.tsx
ChrisChrisLoLo: Good idea. I'm planning to add this in my next PR when I work on those related changes

front-end/src/components/GithubFeed.tsx
ChrisChrisLoLo: I could do this. Currently, I just have logic to make sure the URL is well-formed before making a request. It doesn't quite overlap with the helper, though I'll be sure to use the helper for HTTP responses going forwards! Thanks for showing me that resource!

PR #205, opened 2021-04-03
Disable prepend to public feed if post is unlisted

jmmxp:
Unlisted posts should not show up in the public feed.


VinnyLuu (approved):
LGTM

PR #206, opened 2021-04-03
Make updating GitHub profile easier

jmmxp:
Now only takes in a username instead of the whole GitHub url. 

image


VinnyLuu (approved):
LGTM

PR #207, opened 2021-04-04
update UI, separate inbox to separate page, remove request list modal

raysarinas:
- Split public feed and inbox tabs into separate pages
- Removed top navbar, and is now on the left side (re-sizing the page to a much smaller size will make it look squishy and bad - not sure if we care to fix this)
- Not sure if those are the icons we want to use on the side/navbar

Feed when logged out:
Screen Shot 2021-04-03 at 7 40 15 PM

Feed when logged in:
Screen Shot 2021-04-03 at 7 37 18 PM

Author page looks bad - but we should get in @ChrisChrisLoLo's GitHub stream changes in before fixing it:
Screen Shot 2021-04-03 at 7 37 44 PM

Inbox page and tabs:
Screen Shot 2021-04-03 at 7 37 35 PM
Screen Shot 2021-04-03 at 7 37 26 PM

Followers/following page:
Screen Shot 2021-04-03 at 7 54 42 PM


front-end/src/App.tsx
```diff
@@ -41,11 +43,14 @@ function App() {
       <LoggedInUserContext.Provider value={loggedInUser}>
       <BrowserRouter>
         <div>
-          <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} />
-          <Container fluid={true}>
-            <Switch>
+          <Container>
+            <Row style={{padding: '5%'}} className="justify-content-md-center">
```
raysarinas: is there a better way to have padding?

front-end/src/components/AppNavBar.tsx
```diff
@@ -27,54 +26,96 @@ interface Props {
...
+    <path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M15 12a3 3 0 11-6 0 3 3 0 016 0z" />
+</svg>);
+const loginIcon = (<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" height="25" width="25" stroke="currentColor">
+<path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M11 16l-4-4m0 0l4-4m-4 4h14m-5 4v1a3 3 0 01-3 3H6a3 3 0 01-3-3V7a3 3 0 013-3h7a3 3 0 013 3v1" />
+</svg>);
+const profileIcon = (<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" height="25" width="25" stroke="currentColor">
+<path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M16 7a4 4 0 11-8 0 4 4 0 018 0zM12 14a7 7 0 00-7 7h14a7 7 0 00-7-7z" />
+</svg> );
+
```
raysarinas: I don't like this lol not sure where we want to put all these icons and stuff

raysarinas: was thinking maybe we can install this package since it includes all the heroicons and whole lot more: https://github.com/react-icons/react-icons

front-end/src/components/AppNavBar.tsx
```diff
@@ -27,54 +26,96 @@ interface Props {
...
-          <NavLink to={`/author/${props.loggedInUser?.authorId}`} className="nav-link">Profile</NavLink>
+          <NavLink to="/settings" className="nav-link">
+            {settingsIcon} Settings
+          </NavLink>
         </NavItem>
         <NavItem onClick={close}>
-          <NavLink to="/" className="nav-link" onClick={() => logOut()}>Log Out</NavLink>
+          <NavLink to="/" className="nav-link" onClick={() => logOut()}>
+            {logoutIcon} Log Out
+          </NavLink>
```
raysarinas: thinking of getting rid of the labels next to the icon and have it so that the labels only appear when you hover over the icon - might make the UI look cleaner?

front-end/src/pages/FollowListPage.tsx
```diff
@@ -5,6 +5,7 @@ import { Row, Col, Card, CardBody, CardTitle, Container, Button, CardText, Nav,
 import AuthorListItem from '../components/AuthorListItem';
 import { Author } from '../types/Author';
 import { UserLogin } from '../types/UserLogin';
+var classNames = require('classnames');
```
raysarinas, line 8: doing this instead of installing `classnames` package because it kept breaking the frontend. maybe someone else can fix it? idk

front-end/src/pages/InboxPage.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [activeTab, setActiveTab] = useState("postsTab");
+  const toggle = (tab: any) => {
+    if (activeTab !== tab) setActiveTab(tab);
+  }
+  const mailIcon = (<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" height="25" width="25">
+      <path strokeLinecap="round" strokeLinejoin="round" stroke="black" strokeWidth={2} d="M3 8l7.89 5.26a2 2 0 002.22 0L21 8M5 19h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v10a2 2 0 002 2z" /></svg>);
+  const peopleIcon = (<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" height="25" width="25">
+      <path strokeLinecap="round" strokeLinejoin="round" stroke="black" strokeWidth={2} d="M12 4.354a4 4 0 110 5.292M15 21H3v-1a6 6 0 0112 0v1zm0 0h6v-1a6 6 0 00-9-5.197M13 7a4 4 0 11-8 0 4 4 0 018 0z" /></svg>);
+  const likesIcon = (<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" height="25" width="25">
+      <path strokeLinecap="round" strokeLinejoin="round" stroke="black" strokeWidth={2} d="M4.318 6.318a4.5 4.5 0 000 6.364L12 20.364l7.682-7.682a4.5 4.5 0 00-6.364-6.364L12 7.636l-1.318-1.318a4.5 4.5 0 00-6.364 0z" /></svg>);
```
raysarinas: again would like to move these somewhere else if possible

raysarinas (commented):
addressed some stuff

front-end/src/pages/InboxPage.tsx
Monica-Uwalaka: I am also trying to figure this out. We have to pay if we want the full package though. Maybe we can make our own module for the icons.

Monica-Uwalaka: I figured how to put all of these in a separate file. Going to push my changes in a bit so you'll see it

front-end/src/App.tsx
ChrisChrisLoLo: I would prefer using `rem` as units rather than percent. IIRC from experience, I get misshapen layouts sometimes when I use %'s, because percentages can look undesirable at certain window sizes.

If the padding looks fine when you scale the window around, this is fine too

front-end/src/components/AppNavBar.tsx
ChrisChrisLoLo: Not opposed to it, feel free to install it if you want :+1:

front-end/src/components/AppNavBar.tsx
ChrisChrisLoLo: Personally, I prefer to have the labels there always, since I suspect that showing the label on hover will shift around the other labels, giving the UI a bit of a shaky effect

This is subjective though, so I trust your judgement with this stuff either way :+1:

front-end/src/pages/FollowListPage.tsx
ChrisChrisLoLo: ~~What's the error with this? I'd be cautious to do something like this since I don't know how webpack handles these `requires`~~

ChrisChrisLoLo (approved):
LGTM, fantastic work :+1:

front-end/src/pages/FollowListPage.tsx
ChrisChrisLoLo: I read up what `classnames` does. If you really want to, you can use `clsx` instead to use the import statement: https://www.npmjs.com/package/clsx. Otherwise, this is probably fine, since I suspect `classnames` uses pre ES6 import/exports

front-end/src/App.tsx
raysarinas: I'll change it to use `rem`

front-end/src/components/AppNavBar.tsx
raysarinas: hmm maybe I'm not saying it right or I'm not understanding what you mean (totally get wanting the labels there though), but I was thinking of using this in particular: https://reactstrap.github.io/components/tooltips/

front-end/src/pages/FollowListPage.tsx
raysarinas: I will take a look at using that!

front-end/src/components/AppNavBar.tsx
ChrisChrisLoLo: I see, I though we were planning to render a `<p>` tag  beside the icon or something. Tooltips look nice!

front-end/src/pages/InboxPage.tsx
raysarinas: I'll take a look - I mentioned in a comment above we can install the `react-icons` module: https://github.com/react-icons/react-icons

and instead of having variables, we could just have something like `<mailIcon />`

PR #208, opened 2021-04-04
added icons to posts

Monica-Uwalaka:
Screenshot (160)


ChrisChrisLoLo (commented):
UI looks good, I like the choices of icons!

Could you see why the Netlify deploy is failing for this PR? I suspect you might need to commit whatever is in `../assets/Icons`, as well as save any new npm install you added to the `package.json` file (to do this, use `npm install --save {NEW_PACKAGE}`) to add the dependency to the `package.json` file.

ChrisChrisLoLo (changes requested):
It also looks like the `package-lock.json` file got cleared. Could you uncommit this change? We should keep this file intact

raysarinas:
> UI looks good, I like the choices of icons!
> 
> Could you see why the Netlify deploy is failing for this PR? I suspect you might need to commit whatever is in `../assets/Icons`, as well as save any new npm install you added to the `package.json` file (to do this, use `npm install --save {NEW_PACKAGE}`) to add the dependency to the `package.json` file.

Netlify is complaining because `../assets/Icons` is missing

PR #210, opened 2021-04-05
update author page, likes feed, ui/icons and removed package

raysarinas:
Removed `react-icons` package - will keep as is with the assets folder/files

Updated author page to only have GitHub feed posts. Links on author pages are now list group items, and added the GitHub icon next to the GitHub link. This is what your own page looks like:
Screen Shot 2021-04-05 at 1 22 07 PM

Other author pages:
Screen Shot 2021-04-05 at 1 22 19 PM

Page with bad GitHub link:
Screen Shot 2021-04-05 at 1 21 52 PM

Made "liked" button to be filled in as red:
Screen Shot 2021-04-05 at 1 30 10 PM

Made like feed items into list group items (adjusted padding, I think it looks a bit cleaner):
Screen Shot 2021-04-05 at 1 31 38 PM


front-end/src/pages/AuthorPage.tsx
```diff
@@ -63,69 +62,67 @@ export default function AuthorPage(props: any) {
...
-          <Link className="text-white" to={{ pathname: `/author/${props.loggedInUser.authorId}/followers` }}><Button>Followers</Button></Link>
-        </CardText>
-        <CardText>
-          <Link className="text-white" to={{ pathname: `/author/${props.loggedInUser.authorId}/following` }}><Button>Following</Button></Link>
-        </CardText>
+        <ListGroupItem tag="a" href={`/author/${props.loggedInUser.authorId}/followers`}>Followers</ListGroupItem>
+        <ListGroupItem tag="a" href={`/author/${props.loggedInUser.authorId}/following`}>Following</ListGroupItem>
       </>)
+    } else {
+      return <FollowRequestButton loggedInUser={props.loggedInUser} currentAuthor={author} isFollower={isFollower} setIsFollower={setIsFollower} />
```
jmmxp, line 80: I don't think this will properly show up for a foreign author because the "authorUrl" variable from above always uses our back-end URL. Can you put a TODO for this? I can do it in a later PR.

front-end/src/pages/AuthorPage.tsx
```diff
@@ -63,69 +62,67 @@ export default function AuthorPage(props: any) {
...
     if (props.loggedInUser && author?.id === props.loggedInUser.authorId) {
       return (<>
-        <CardText>
-          <Link className="text-white" to={{ pathname: `/author/${props.loggedInUser.authorId}/followers` }}><Button>Followers</Button></Link>
-        </CardText>
-        <CardText>
-          <Link className="text-white" to={{ pathname: `/author/${props.loggedInUser.authorId}/following` }}><Button>Following</Button></Link>
-        </CardText>
+        <ListGroupItem tag="a" href={`/author/${props.loggedInUser.authorId}/followers`}>Followers</ListGroupItem>
+        <ListGroupItem tag="a" href={`/author/${props.loggedInUser.authorId}/following`}>Following</ListGroupItem>
```
jmmxp: We already discussed this offline, but maybe just having one button "Friends" or "Follows" might look better.

jmmxp (approved):
LGTM just some comments to address

Also, awesome work here, everything looks really good.

front-end/src/pages/AuthorPage.tsx
jmmxp: Just to clarify, the authorUrl variable always uses our back-end, making it so that the `author` object used here would break if it's not from our back-end.

PR #212, opened 2021-04-06
Fix comment race condition

jmmxp:
Commenting sometimes didn't work, because we were checking commentAuthor state in an if statement, but setCommentAuthor is 
presumably asynchronous. I've fixed this by now returning the author from the Promise so that we can use it as a parameter to .then().
 
image


ChrisChrisLoLo (approved):
Good catch, LGTM :+1:

PR #213, opened 2021-04-06
Return newly created comments from back-end and prepend to feed in front-end

jmmxp:
I've updated the CommentViewSet to return newly created comments, so that the front-end can retrieve this newly created comment (with its ID) and prepend it to the comment list feed. Previously this wasn't doable as pointed out by @raysarinas because liking comments would fail without a proper ID.

### DIFF
--- a/front-end/src/components/CommentList.tsx
+++ b/front-end/src/components/CommentList.tsx
@@ -14,7 +14,7 @@ interface Props {
   postAuthor: Author,
   loggedInUser: UserLogin | undefined,
   setUpdateComment: Function,
-  updateComment: boolean,
+  updateComment: PostComment | undefined,
 }
 
 /**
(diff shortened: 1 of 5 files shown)


front-end/src/components/CommentList.tsx
```diff
@@ -1,54 +1,20 @@
...
-import { Author } from '../types/Author';
+import { Col, Container } from 'reactstrap';
 import { PostComment } from '../types/PostComment';
 import CommentListItem from './CommentListItem';
-import { AxiosWrapper } from '../helpers/AxiosWrapper';
 
 interface Props {
-  postId: string,
-  postAuthor: Author,
-  loggedInUser: UserLogin | undefined,
```
VinnyLuu: Can you leave this in. I'm going to need these to pass to CommentListElement for liking comments.

VinnyLuu (approved):
LGTM

jmmxp:
Thanks to @ChrisChrisLoLo for giving me a tentative code review too to try and figure out why I was having infinite re-render issues.

front-end/src/components/CommentList.tsx
jmmxp: Done.

PR #214, opened 2021-04-06
Fixed up likes for comments

VinnyLuu:
Added trailing slash to like comment endpoint. 

Queries for likes of a post are polluted by likes for comments on that post because the object field for those comments will also include the post id. This pr filters out any likes out that are from comments by excluding any Like that contains the substring comment.


jmmxp (approved):
LGTM

PR #216, opened 2021-04-07
Fixed slow follow button text switching

VinnyLuu:
Changed so that now when clicking the follow button it will immediately switch, only switching back when there is an error.


jmmxp (approved):
LGTM

PR #218, opened 2021-04-07
Add timestamps to comments

jmmxp:
Updated the getDateString method to take in any object. Now using this method to show timestamps for comments.

image


VinnyLuu (approved):
LGTM

PR #219, opened 2021-04-07
Allow title and description to be blank when editing post

jmmxp:
This will give more freedom to users when editing a post. Currently we have a bug where editing a post and having a blank description is not allowed. Warning: these blank=True / blank=False is for some reason ignored when creating a post - by default it seems that creating a post is OK with everything being blank.

Note that I would prefer to have title not blank, but this will result in a lot of changes in the back-end. We should just sanity check inputs on the front-end.


VinnyLuu (approved):
LGTM

PR #220, opened 2021-04-07
Update AuthorPage to work with cross-server authors

jmmxp:
Keeping track of author host upon login/registration. Now when we navigate to the profile page (either by the search bar, or by clicking profile page), we send in author host+id so that we can make a request that also allows for cross server.


VinnyLuu (approved):
LGTM

PR #221, opened 2021-04-08
Implemented comment pagination

VinnyLuu:
Comments are now paginated. The default number of comments for a page is 5, but can be changed with the query parameters

An example on how to use the paginated comments:
/api/author/{AUTHOR_ID}/posts/{POST_ID}/comments/?size=10&page=2
where size specifies the number of elements for a page, and page specifies which page to get.


jmmxp (approved):
LGTM very nice

PR #222, opened 2021-04-08
Allow display name and github username on registration

jmmxp:
Display name is required but Github username is optional.

Not putting a GitHub username results in it being set to your username. (as done in signals.py)

image


VinnyLuu (approved):
LGTM

PR #223, opened 2021-04-08
Re-authenticate GET Inbox and singular Post endpoint

jmmxp:
We removed auth from the GET Inbox endpoint because we had bugs with AxiosWrapper on the front-end and the Friend Requests tab that was in the app nav bar. AxiosWrapper has since been updated and the FriendRequests tab has been moved into the Inbox, which re-fetches every single time you click it, so re-authing this should be safe.

These two endpoints are authed as per the spec.


VinnyLuu (approved):
LGTM

PR #224, opened 2021-04-08
Don't make requests to comments if we aren't logged in

jmmxp:
Making unnecessary requests when loading public feed.


VinnyLuu (approved):
LGTM

PR #225, opened 2021-04-08
Frontend for comment pagination

VinnyLuu:
Similar to reddit, below the comments will be a button to show more comments. Clicking on the button will append the new list of paginated comments to the comment list. When there are no more comments, the button to get more comments will disappear.

Show more comments button:
image

No more comments:
image


jmmxp (approved):
LGTM

PR #227, opened 2021-04-08
Show more comments button properly hidden if no more comments

VinnyLuu:
Fix that if no comments exist at first or there is no more than 5, no comments button doesn't show up


jmmxp (approved):
LGTM

PR #229, opened 2021-04-09
Fix crash when navigating to profile page from followers/following list

jmmxp:
Needed to send in clicked author's host and ID so that the profile page can make a proper request to get that author's data.


raysarinas (approved):
LGTM

PR #230, opened 2021-04-09
Even more UI updates

raysarinas:
Show posts on your own profile now, and we only make a get request to get the posts when you click the tab. Show GitHub feed on default since its also the default state for other author profiles that aren't your own. Also added the pencil icon to the create post button
Screen Shot 2021-04-09 at 3 31 01 PM
Screen Shot 2021-04-09 at 3 32 02 PM

Other author profile:
Screen Shot 2021-04-09 at 3 32 20 PM

Moved clear inbox to be inline:
Screen Shot 2021-04-09 at 3 32 36 PM

Hide the search bar if logged out:
Screen Shot 2021-04-09 at 3 33 05 PM


VinnyLuu (approved):
Awesome LGTM!!

PR #232, opened 2021-04-11
send post to post_id

Monica-Uwalaka:


front-end/package.json
```diff
@@ -18,6 +18,7 @@
     "dateformat": "^4.5.1",
     "node-sass": "^4.14.1",
     "react": "^17.0.1",
+    "react-bootstrap": "^1.5.2",
```
raysarinas: Is there a reason you installed react-bootstrap? Doesn't look like it's being used anywhere. We're using reactstrap and having react-bootstrap might cause problems

front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -7,6 +7,10 @@ import { UserLogin } from "../types/UserLogin";
 import { Author } from "../types/Author";
 import { ResponseHelper } from "../helpers/ResponseHelper"
 import PostContentEl from "./PostContentEl";
+import * as Style from '../assets/createPostUI';
```
raysarinas, line 10: `createPostUI` file was not pushed to Git and is causing the checks to fail on netlify

front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -117,15 +133,32 @@ export default function CreateEditPostModal(props: Props){
           }).catch((error: any) => {
             setShowError(true)
           })
+
+        //put to post_id
+        const uuid = generateUUID();
+        console.log(uuid)
+        AxiosWrapper.put(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/" + uuid + "/", data, props.loggedInUser)
```
raysarinas: please use a template string here instead of concatenating strings. For example, `${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/posts/${uuid}/`

front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -117,15 +133,32 @@ export default function CreateEditPostModal(props: Props){
...
           }).catch((err: any) => {
             setShowError(true)
           })
+
+          //put to post_id
+          const uuid = generateUUID();
+          AxiosWrapper.put(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/" + uuid + "/", data, props.loggedInUser)
+          .then((res: any) => {
+            handleRes(res)
+          })
```
raysarinas: maybe there's something in the missing file, but we should only put to post_id when a user clicks a button to generate a UUID. if the user does not click the button, the uuid field should be empty and this put request should not be called.

raysarinas (changes requested):
please add missing files and we should discuss more during the meeting

front-end/package.json
Monica-Uwalaka: yes I was trying to do something with the UI but I ended up not using it. I'll remove it.

front-end/src/components/CreateEditPostModal.tsx
Monica-Uwalaka: I missed that

front-end/package.json
Monica-Uwalaka: Actually we need it for App.scss

jmmxp: App.scss isn't used so it seems like you can delete it.

front-end/src/components/CreateEditPostModal.tsx
```diff
@@ -99,9 +110,16 @@ export default function CreateEditPostModal(props: Props){
...
-        AxiosWrapper.post(process.env.REACT_APP_API_URL + "/api/author/" + props.loggedInUser.authorId + "/posts/", data, props.loggedInUser)
-          .then((res: any) => {
+        if (uuid === ""){
+          promise = AxiosWrapper.post(`${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/posts/`, data, props.loggedInUser);
+        }
+          else{
+            promise = AxiosWrapper.put(`${process.env.REACT_APP_API_URL}/api/author/${props.loggedInUser.authorId}/posts/${uuid}/`, data, props.loggedInUser)
+          }
+          
+          promise.then((res: any) => {
```
jmmxp: Indentation of this promise and everything below looks like it's 1 tab off

jmmxp (approved):
LGTM, package-lock.json files might have some issues but we'll double check after merging.

PR #233, opened 2021-04-11
Make GitHub feed show create PR events only, update UI

jmmxp:
GitHub feed only gets PullRequestEvents with action: opened now. I also changed the request for events to include page_size of 100, which is the max (https://docs.github.com/en/rest/reference/activity#events). Comparison of the UI below.

**Before:**
eventsbefore

**After:**
eventsafter

### DIFF
--- a/front-end/src/components/GithubFeedItem.tsx
+++ b/front-end/src/components/GithubFeedItem.tsx
@@ -9,7 +9,7 @@ interface Props {
  * github list event component to show the author's github events
  * @param props
  */
-export default function LikeListItem(props: Props) {
+export default function GithubFeedItem(props: Props) {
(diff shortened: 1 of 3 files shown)


front-end/src/components/GithubFeedItem.tsx
```diff
@@ -10,11 +10,16 @@ interface Props {
  * @param props
  */
 export default function LikeListItem(props: Props) {
```
raysarinas: I think this should be `GithubFeedItem`?

jmmxp: Fixed

raysarinas (approved):
LGTM

PR #234, opened 2021-04-11
Show alert upon successful reshare

jmmxp:
When the post is successfully created I immediately show the alert. Note that we can move the alert to after we send it to followers' inboxes but this would make the alert quite slow.

image


VinnyLuu (approved):
LGTM

PR #235, opened 2021-04-11
Added checks if follow/following exists before creating

VinnyLuu:
Previously, django would complain about having multiple instances of a follow/following if they were PUT multiple times. Here I added checks to make sure that multiple instances are not created and instead just return a 204 if it already exists.


ChrisChrisLoLo (approved):
LGTM!
Are changes in the FE needed to catch this error response?

PR #236, opened 2021-04-11
Checked id key exists before iterating for friends

VinnyLuu:
Faulty id fields cause a key exception for the friends endpoint. Added in a check to make sure that the key exists.

### DIFF
--- a/back-end/api/quickstart/views.py
+++ b/back-end/api/quickstart/views.py
@@ -202,7 +202,7 @@ class FriendsListViewSet(viewsets.ModelViewSet):
             friends = []
             for follow in follow_queryset:
                 for following in following_queryset:
-                    if follow.sender["id"] == following.receiver["id"]:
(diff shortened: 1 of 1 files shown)


jmmxp (approved):
LGTM

jmmxp:
are there any other places we're trying to read fields from JSONFields? we should make sure we have those checks there too

ChrisChrisLoLo (approved):
Nice! I'm thinking this should help, though it may be possible that some error get's propagated (for example, the error won't be sent here, but the FE might break somewhere down the road). 

It might be good to test the inboxes while following someone in anas's server. If I get some time I'll do some testing tomorrow

ChrisChrisLoLo:
> are there any other places we're trying to read fields from JSONFields? we should make sure we have those checks there too

Do you think we put ourselves at obfuscating the error if we do this? Even if the server doesn't throw an error now, do you think the FE could break somewhere down the road regardless?

PR #237, opened 2021-04-12
Changed url building for comments to use host instead of url

VinnyLuu:
This change makes it consistent with how urls are built elsewhere. Also Anas's team doesn't seem to be sending url in author objects so it breaks xserver as it will be undefined.


jmmxp (approved):
LGTM

PR #240, opened 2021-04-12
Even more UI updates

raysarinas:
Author page looks like this now:
- followers/following are tabs and got rid of following list page
- cleaned up some GitHub feed item stuff since it was looking weird on my end but its basically the same
- changed how author list items look like. clicking on their names will link to their host or profile; changed follow/unfollow buttons to icons
Screen Shot 2021-04-11 at 10 35 53 PM

Fixed a weird bug with GitHub profile pics so now it always goes to this default if either there is no valid GitHub url OR if you have one initially, then clear it and all that remains is "https://github.com/" without a username
Screen Shot 2021-04-11 at 10 36 04 PM

Janky fix done to hide the create post button on the inbox and your own profile page:
Screen Shot 2021-04-11 at 10 54 20 PM


raysarinas:
more changes:

- added logo to navbar, changed post list items and right-aligned buttons:
Screen Shot 2021-04-12 at 4 58 07 PM

- added logo to login page and hide all pages if not logged in. will redirect to 404:
Screen Shot 2021-04-12 at 4 58 22 PM

- 404 page:
Screen Shot 2021-04-12 at 4 58 38 PM

front-end/src/App.tsx
```diff
@@ -44,20 +43,23 @@ function App() {
...
             <Row style={{padding: '3rem'}} className="justify-content-md-center">
-              <Col xs="2"><AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} /></Col>
+              <Col md="2">
+                <AppNavBar loggedInUser={loggedInUser} setLoggedInUser={setLoggedInUser} /></Col>
               <Col>
               <Switch>
               <Route exact path="/" render={(props) => <HomePage {...props} loggedInUser={loggedInUser} />} />
-              <Route path="/auth" render={(props) => <AuthPage {...props} setLoggedInUser={setLoggedInUser} />} />
+              <Route path="/login" render={(props) => <AuthPage {...props} setLoggedInUser={setLoggedInUser} />} />
+              {loggedInUser &&
```
ChrisChrisLoLo, line 52: I'm not sure if this could cause bugs for some features (links and stuff).
Here's how to do React Router Auth redirects, though it may take more time than what we have left. https://reactrouter.com/web/example/auth-workflow
Personally, I'd rather go without this change since our current workflow and demo work is alright

front-end/src/assets/Icons.tsx
```diff
@@ -4,6 +4,19 @@ interface StyleProps{
...
+	<path d="M20,2c2.071,0,3.933,0.569,5.905,1.171c0.739,0.226,1.503,0.459,2.307,0.673   c0.599,0.159,1.229,0.299,1.838,0.435c2.045,0.455,4.159,0.925,5.23,2.148C38,9.536,38,15.584,38,20   c0,5.519-1.982,7.781-4.725,10.911l-0.302,0.345c-0.599,0.685-1.098,1.328-1.581,1.95C29.151,36.093,27.671,38,20,38   c-6.946,0-9.604-3.34-11.192-5.336c-0.327-0.411-0.61-0.766-0.892-1.049C6.928,30.627,2,25.424,2,20   c0-3.247,3.047-5.545,5.735-7.573c1.306-0.985,2.433-1.836,3.137-2.741c0.623-0.801,0.965-1.783,1.295-2.732   C13.054,4.406,13.892,2,20,2 M20,1c-9.195,0-7.896,5.473-9.917,8.072C8.062,11.67,1,14.587,1,20c0,5.567,4.679,10.793,6.209,12.322   C8.739,33.852,11.175,39,20,39c9.789,0,10.243-3.101,13.726-7.085C36.646,28.574,39,26.107,39,20c0-4.788-0.044-10.89-2.967-14.231   c-1.643-1.878-4.897-2.182-7.564-2.891C25.483,2.083,23.014,1,20,1L20,1z"/>
```
ChrisChrisLoLo, line 10: Nice! Thanks for making this SVG!

front-end/src/components/AppNavBar.tsx
```diff
@@ -63,7 +70,7 @@ export default function AppNavBar(props: Props) {
           </NavLink>
         </NavItem>
         <NavItem onClick={close}>
-          <NavLink to="/" className="nav-link" onClick={() => logOut()}>
+          <NavLink to="/login" className="nav-link" onClick={() => logOut()}>
```
ChrisChrisLoLo, line 73: :+1:

front-end/src/components/AppNavBar.tsx
```diff
@@ -73,30 +80,21 @@ export default function AppNavBar(props: Props) {
...
-                <NavLink to="/auth/" className="nav-link">
-                {Icons.loginIcon} Log In
-                </NavLink>
-              </NavItem>
-            )}
+            {loggedIn()}
           </Nav>
         </Collapse>
       </Navbar>
+      : <h1 style={{padding: '4rem'}}>{Icons.tamagoEgg("150", "150")}</h1>
```
ChrisChrisLoLo: Do we have a way to navigate to the login page if we are not logged in?

front-end/src/components/FriendRequestButton.tsx
```diff
@@ -53,7 +54,7 @@ export default function FollowRequestButton(props: Props) {
 
   return (
     <>
-      <Button className="list-group-item" color="link" onClick={sendFollowRequest}>{props.isFollower ? "Unfollow" : "Follow"}</Button>
+      <Button style={{outline: 'none'}} color="white" onClick={sendFollowRequest}>{props.isFollower ? Icons.unfollowIcon : Icons.followIcon}</Button>
```
ChrisChrisLoLo: Nit: As a fun fact, my mentor told me in production one should never set `outline` to `none`, since it affects accessibility (makes it harder to navigate via keyboard).
https://stackoverflow.com/questions/19455499/why-is-outlinenone-not-considered-good-practice

This is a fun fact only, I'm more than fine leaving it in for our usecase

front-end/src/pages/AuthorPage.tsx
```diff
@@ -99,31 +120,31 @@ export default function AuthorPage(props: any) {
 
 
   const displayProfileButtons = () => {
-    if (props.loggedInUser && author?.id === props.loggedInUser.authorId) {
-      return <ListGroupItem tag="a" href={`/author/${props.loggedInUser.authorId}/followers`}>Follows</ListGroupItem>
-    } else {
+    if (author?.id !== props.loggedInUser.authorId) {
```
ChrisChrisLoLo: what if `props.loggedInUser` is undefined?

front-end/src/pages/AuthorPage.tsx
```diff
@@ -50,13 +51,33 @@ export default function AuthorPage(props: any) {
   const [responseMessage, setResponseMessage] = useState(100);
```
ChrisChrisLoLo, line 52: On line 47 we should bind the props of this to type `Props`. This is a bit of scope creep. However, since we're changing around the props in this component, I think it would be good to set the prop type to be confident that we're not introducing any bugs

ChrisChrisLoLo (commented):
I think the UI changes made here are great! :+1:

My biggest concern with this PR is that there seems to be a lot of logic being changed around. While you have a very good record of making code free of defects, the coverage of these changes makes it possible that something could be slipping through the cracks near the deadline.

We have a few options going forward:
- Address the prop type comment, and merge as is. We can test tomorrow, however, if there's a snag in the demo, we may need you or someone else to hotfix these changes
- Revert some of the logic in this change (mostly with respect to authentication and redirection).

I don't want to undo work that you've put time and care into, so I'm biasing towards the first option, with the understanding that someone is available to test and fix bugs tomorrow if something goes awry. What are your thoughts?

front-end/src/App.tsx
raysarinas: could u point to me what change I should revert? Honestly if any of us doesn't like something or is unsure of any potential issues I think we should just scrape it

front-end/src/components/AppNavBar.tsx
raysarinas: The changes I made always redirect to the login page if you're not logged in. Otherwise if you try to navigate anywhere other than the login page you'll get a 404

front-end/src/components/FriendRequestButton.tsx
raysarinas: I'll change this. what color should the button be? might look a lil wonky which is my concern

front-end/src/pages/AuthorPage.tsx
raysarinas: good catch, I'm like 50% sure the if statement fails and then returns nothing (see line 147)

front-end/src/pages/AuthorPage.tsx
raysarinas: agreed, I think the reason why some places we have `props: any` is because like idk how to explain it but there are issues that come up with typing some of the props that get passed in

front-end/src/components/FriendRequestButton.tsx
ChrisChrisLoLo: Your pick! I only brought it up since you seem fairly invested in FE stuff, so I'd like to pass on anything I know. If it looks wonky we can leave it as is

front-end/src/pages/AuthorPage.tsx
ChrisChrisLoLo: I think the `FollowListPage` component has the typing properly implemented. I want to be a bit firmer with this alone since if we can type the props we can instantly find any `undefined` bugs if they exist

ChrisChrisLoLo: If it doesn't work out maybe make an issue and tag me in it so I can take a look. I don't want to block you too long on this considering the timeline

front-end/src/pages/AuthorPage.tsx
ChrisChrisLoLo: if `props.loggedInUser` is undefined then IIRC we might be at risk of an undefined error such as `Error: Cannot read property 'authorId' of undefined`. If we can get typing on the incoming props then TS should be able to point these cases out to us for free

front-end/src/components/AppNavBar.tsx
ChrisChrisLoLo: Thanks! While there's minor UX trade-offs with this approach (can't press the back button while in the login screen), I think this implementation should be good for our project :+1:

front-end/src/App.tsx
ChrisChrisLoLo: After thinking about it and playing around with it a bit your implementation looks good for our demo :rocket: . If we find anything we can catch it during our test run tomorrow

ChrisChrisLoLo (approved):
After thinking about it a bit I think we should be good to merge this in :shipit: 
I would recommend either looking at typing some of the page props or marking me down to take a look at it, since it can help us catch some edge cases with `undefined`s.

After thinking about it for a bit, as well as demoing it via Netlify, there isn't any obvious regressions, and any small thing that could come up should be fixable tomorrow.

Thanks for making this PR to polish the UI! It's very clear that thought and effort were put into it.

front-end/src/components/AppNavBar.tsx
raysarinas: maybe I misunderstood but you can click the egg to go back to the login screen? Like if you get a 404, there's a button to go to login and you could also click the egg

raysarinas: wait I forgot to push the change I made so you can click the egg to go back to the login screen lol I'll do that now

front-end/src/components/FriendRequestButton.tsx
raysarinas: I took out the style/outline stuff, doesn't seem to have changed anything so we're good

front-end/src/pages/AuthorPage.tsx
raysarinas: fixed (sort of)

front-end/src/pages/AuthorPage.tsx
raysarinas: I'll make an issue, I know that the problem is that I think when we were working on the AuthorPage, at the time we didn't know how to type `props.location`. That is causing issues

ChrisChrisLoLo: We can type `.location` by doing what we have in `FollowListPage`, which should extend the React Router Props interface type. The problem might be more complicated than that, but that'll be a good first place to start

ChrisChrisLoLo: I'll take a quick look to see if I can type it in your branch. Feel free to merge whenever though

front-end/src/components/AppNavBar.tsx
ChrisChrisLoLo: If I get redirected to the login screen, I can't press the back button, as I get sent back to the page that will redirect me back to the login screen. React router handles these kinds of cases with it's auth features, but we don't need to worry about it for this project

PR #241, opened 2021-04-12
Fixed 204 handling for comments

VinnyLuu:
Fixed backend to send 204 when no comments exist for the page. Also fixed frontend to properly handle 204 if the post is new. When trying to add a comment on a new post with no comments, commentslist is undefined, so when adding a comment, it is not properly appended.


jmmxp (approved):
LGTM

